Issue #1, opened 2019-10-07
ICE: trying to add a local var with the same name, but different types. during [_RegisterClipboardFormat]


ams-tschoening:
I have a PoC to use some existing Java-codebase in some UWP-app using the most current Visual Studio Community 19 version 16.3.2 and the latest released IKVM 8.1.7195.0. The app builds and runs fine in Debug-mode, but fails to build already in Release-mode with the following error:

	Schweregrad	Code	Beschreibung	Projekt	Datei	Zeile	Unterdrückungszustand
	Fehler		MCG0004:InternalAssert Assert Failed: ICE: trying to add a local var with the same name, but different types. during [_RegisterClipboardFormat]	Ams.Oms.Poc			

`RegisterClipboardFormat` is part of IKVM:

    @DllImportAttribute.Annotation(value = "user32.dll", EntryPoint = "RegisterClipboardFormat")
    private native static int _RegisterClipboardFormat(String format);

    @cli.System.Security.SecuritySafeCriticalAttribute.Annotation
    private static int RegisterClipboardFormat(String format)
    {
        return _RegisterClipboardFormat(format);
    }

https://github.com/ikvm-revived/ikvm/blob/master/openjdk/sun/awt/IkvmDataTransferer.java#L95

What I'm wondering is which local variable the error message is referring to? Might be something added implicitly or might have to do with `String` in Java vs. `string` in C#? OTOH that file is clearly named `.java`.

Didn't find much about the error message in general, only the following two links seems to be most interesting:

https://stackoverflow.com/questions/13657026/variables-having-same-name-but-different-type
https://stackoverflow.com/questions/4649947/why-doesnt-c-sharp-allow-me-to-use-the-same-variable-name-in-different-scopes

So I'm currently even unsure where the message comes from, Visual Studio/C# directly or IKVM during running code during building Release-mode already. I strongly suspect the error coming from C#, though.

Searching for the function itself doesn't reveal much of help as well:

> Sorry, AWT is not a supported part of IKVM.

https://sourceforge.net/p/ikvm/bugs/225/

Others seem to have the same problem, because CN1 simply disabled that code entirely in their fork of IKVM:

    //@DllImportAttribute.Annotation(value = "user32.dll", EntryPoint = "RegisterClipboardFormat")
    //private native static int _RegisterClipboardFormat(String format);

    @cli.System.Security.SecuritySafeCriticalAttribute.Annotation
    private static int RegisterClipboardFormat(String format)
    {
[...]

PR #2, opened 2019-10-24
Workaround for internal compiler error regarding some native functions.

ams-tschoening:
The function commented in this commit led errors like the following during compiling a Release-version of some UWP-app already:

> ICE: trying to add a local var with the same name, but different types. during [_RegisterClipboardFormat]

The CN1-fork simply commented those functions as well, so I followed that approach for now. The important thing is that I really only commented those functions which triggered the error until that error didn't occur anymore. So I did not simply commented all native functions by purpose, because currently it seems that some trigger that error, while some don't.

The problem is that currently another error is triggered during compilation, so I can't be sure if that is fixed somehow, the error of this commit happens again with the other functions not yet commented. Because of the second error, committing the current changes seems the best way to go to keep things independent.

This is related to https://github.com/ikvm-revived/ikvm/issues/1.

### DIFF
--- a/openjdk/java/awt/Font.java
+++ b/openjdk/java/awt/Font.java
@@ -821,12 +821,12 @@ public class Font implements java.io.Serializable
     }
 
     private static class FontFamilyReference extends WeakReference<FontFamily> {
-        
+
         private static final ReferenceQueue<FontFamily> fontFamilyQueue = new ReferenceQueue<>();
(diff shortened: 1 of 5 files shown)


ams-tschoening:
This is only the workaround implemented by CN1 and doesn't fix #1, so should be handled with care. I'm providing it for documentation reasons and maybe someone see what the currently commented function calls have in common and might trigger the root cause of this issue. I find it pretty surprising that at least up to the point where I get different compiler errors, not all native functions seems to be affected.

PR #4, opened 2019-10-24
Workaround for internal compiler error regarding some native function…

ams-tschoening:
… calls.

The function commented in this commit led errors like the following during compiling a Release-version of some UWP-app already:

> ICE: trying to add a local var with the same name, but different types. during [_RegisterClipboardFormat]

The CN1-fork simply commented those functions as well, so I followed that approach for now. The important thing is that I really only commented those functions which triggered the error until that error didn't occur anymore. So I did not simply commented all native functions by purpose, because currently it seems that some trigger that error, while some don't.

The problem is that currently another error is triggered during compilation, so I can't be sure if that is fixed somehow, the error of this commit happens again with the other functions not yet commented. Because of the second error, committing the current changes seems the best way to go to keep things independent.

This is related to https://github.com/ikvm-revived/ikvm/issues/1.

### DIFF
--- a/openjdk/java/awt/Font.java
+++ b/openjdk/java/awt/Font.java
@@ -821,12 +821,12 @@ public class Font implements java.io.Serializable
     }
 
     private static class FontFamilyReference extends WeakReference<FontFamily> {
-        
+
         private static final ReferenceQueue<FontFamily> fontFamilyQueue = new ReferenceQueue<>();
         private static final Set<FontFamilyReference> refs = Collections.synchronizedSet( new HashSet<FontFamilyReference>() );
         static {
-            sun.misc.SharedSecrets.getJavaLangAccess().registerShutdownHook( 5, // Shutdown hook invocation order 
-            true, // register even if shutdown in progress 
+            sun.misc.SharedSecrets.getJavaLangAccess().registerShutdownHook( 5, // Shutdown hook invocation order
+            true, // register even if shutdown in progress
             new Runnable() {
                 public void run() {
                     for( FontFamilyReference ref : refs ) {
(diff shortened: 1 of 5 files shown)


ams-tschoening:
This is only the workaround implemented by CN1 and doesn't fix #1, so should be handled with care. I'm providing it for documentation reasons and maybe someone see what the currently commented function calls have in common and might trigger the root cause of this issue. I find it pretty surprising that at least up to the point where I get different compiler errors, not all native functions seems to be affected.

DavidThi808:
I didn't look at the awt parts. But the file I/O changes look good to me. So I approve this merge.

DavidThi808:
My one concern is - are these methods needed? If so, we should do an #if NETFRAMEWORK around the old way.

ams-tschoening:
@DavidThi808 I've created an additional PR #16 for the available file-I/O-stuff. Let's not mix it with this and the other PR #5, those are different. I don't know the reasons for those problems yet, so I suggest ignoring them for now in favour of the other work which is known as needed. Maybe the other work fixes these problems as well.

ams-tschoening:
`#if NETFRAMEWORK` doesn't work in Java like mentioned in [#16](https://github.com/ikvm-revived/ikvm/pull/16#discussion_r389381735). Checking things on runtime for the current version of .NET executed under [reads complicated](https://stackoverflow.com/a/51974946/2055163) to me.

One should focus on getting to know about the underlying root cause if the problem first before implementing different things. The PR is a workaround(!) only to get things to run at all.

ams-tschoening:
Closing this for now because of the heavy changes [mentioned](https://github.com/ikvm-revived/ikvm/issues/13#issuecomment-632030448) and [implemented](https://github.com/klokedm/ikvm) regarding .NET Core compatibility WITHOUT these workarounds.

This is to make it easier for people to follow what's going on at all and because it's currently not known if these workaround are needed or not. In bets case they are not, but if people run into the same problems again, this PR can simply be reopened and is not lost.

Just DON'T delete the associated branch too easily! :-)

Issue #3, opened 2019-10-24
error CS0100: The parameter name '__value' is a duplicate"


ams-tschoening:
Not sure if this is related to #1 or something completely different, but after applying PR #2 I get an additional, but new error:

    Schweregrad	Code	Beschreibung	Projekt	Datei	Zeile	Unterdrückungszustand
    Fehler		ILT0014: Fehler beim Kompilieren des Interop-Quellcodes. Fehlerdetails finden Sie im Buildprotokoll.	Ams.Oms.Poc			

Some log file `obj\x86\Release\ilc\ilclog.csv` contains additional details:

    "Message",Normal,"obj\x86\Release\ilc\intermediate\Ams.Oms.Poc.McgInterop\PInvoke.g.cs(3649,10): error CS0100: The parameter name '__value' is a duplicate"

The log contains multiple such statements, because the mentioned file contains multiple errors like the following:

	// Signature, MapViewOfFile, [fwd] [return] [EETypeRva:0x000419B0] System_IntPtr____w64 int, [fwd] [in] [EETypeRva:0x000434A4] Microsoft_Win32_SafeHandles_SafeFileHandle__System_Private_CoreLib____w64 int, [fwd] [in] [EETypeRva:0x000419B0] int__int, [fwd] [in] [EETypeRva:0x000419B0] int__int, [fwd] [in] [EETypeRva:0x000419B0] int__int, [fwd] [in] [EETypeRva:0x000419B0] System_IntPtr____w64 int, 
	[global::System.Runtime.InteropServices.McgPInvokeMarshalStub("IKVM.OpenJDK.Core, Version=8.1.7234.0, Culture=neutral, PublicKeyToken=null", "sun.nio.ch.FileChannelImpl", "MapViewOfFile")]
	public static global::System.IntPtr MapViewOfFile(
				global::Microsoft.Win32.SafeHandles.SafeFileHandle__System_Private_CoreLib __value, 
				int __value, 
				int __value, 
				int __value, 
				global::System.IntPtr __value)

Using the same name `__value` multiple times clearly doesn't make much sense. This error is interesting because the generated code perfectly well fits to what the error message of #1 says and the function in this case is a native function which I didn't need to comment because it didn't trigger the error of #1, even though it looks pretty the same:

    @DllImportAttribute.Annotation(value = "user32.dll", EntryPoint = "RegisterClipboardFormat")
    private native static int _RegisterClipboardFormat(String format);

vs.

[...]

PR #5, opened 2019-10-24
Workaround for code generating multiple `__value`-args.

ams-tschoening:
The functions commented in this commit led to errors like the following during compiling a Release-version of some UWP-app:

> error CS0100: The parameter name '__value' is a duplicate"

The CN1-fork simply commented those functions as well, so I followed that approach for now. The important thing is that I really only commented those functions which triggered the error until that error didn't occur anymore. So I did not simply commented all native functions by purpose, because currently it seems that some trigger that error, while some don't. The difference most likely simply is the number of arguments, because one "__value" is not a problem, while two are.

This is related to https://github.com/ikvm-revived/ikvm/issues/3.

### DIFF
--- a/openjdk/sun/nio/ch/FileChannelImpl.java
+++ b/openjdk/sun/nio/ch/FileChannelImpl.java
@@ -1032,7 +1032,7 @@ public class FileChannelImpl
             int PAGE_READONLY = 2;
             int PAGE_READWRITE = 4;
             int PAGE_WRITECOPY = 8;
-            
+
             int FILE_MAP_WRITE = 2;
             int FILE_MAP_READ = 4;
             int FILE_MAP_COPY = 1;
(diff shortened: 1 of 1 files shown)


DavidThi808:
My one concern is - are these methods needed? If so, we should do an #if NETFRAMEWORK around the old way.

ams-tschoening:
Someone might use them of course and things break then. They are no real solutions, someone else needs to try to understand the underlying problem. I simply didn't. :-)

ams-tschoening:
`#if NETFRAMEWORK` doesn't work in Java like mentioned in [#16](https://github.com/ikvm-revived/ikvm/pull/16#discussion_r389381735). Checking things on runtime for the current version of .NET executed under [reads complicated](https://stackoverflow.com/a/51974946/2055163) to me.

One should focus on getting to know about the underlying root cause if the problem first before implementing different things. The PR is a workaround(!) only to get things to run at all.

ams-tschoening:
Closing this for now because of the heavy changes [mentioned](https://github.com/ikvm-revived/ikvm/issues/13#issuecomment-632030448) and [implemented](https://github.com/klokedm/ikvm) regarding .NET Core compatibility WITHOUT these workarounds.

This is to make it easier for people to follow what's going on at all and because it's currently not known if these workaround are needed or not. In bets case they are not, but if people run into the same problems again, this PR can simply be reopened and is not lost.

Just DON'T delete the associated branch too easily! :-)

PR #16, opened 2020-03-02
Net core compat to master.

ams-tschoening:
This PR merges the branch `net_core_compat` to master and can be reviewed individually.

### DIFF
--- a/openjdk/sun/nio/fs/NetFileSystemProvider.java
+++ b/openjdk/sun/nio/fs/NetFileSystemProvider.java
@@ -372,24 +372,15 @@ final class NetFileSystemProvider extends AbstractFileSystemProvider
             if (false) throw new cli.System.IO.IOException();
             if (false) throw new cli.System.Security.SecurityException();
             if (false) throw new cli.System.UnauthorizedAccessException();
-
             // TODO NET_CORE_INCOMPAT
-            //
-            // Some callers expect atomic appends, but those are not possible currently because of
-            // a missing CTOR, but doesn't mean that we can't allow all other requests.
+            // TODO NET_STANDARD_INCOMPAT
             //
             // https://github.com/dotnet/corefx/issues/39920
-            // https://github.com/wwrd/ikvm8/issues/3#issuecomment-517009923
-            //
-            // https://stackoverflow.com/questions/1862309/how-can-i-do-an-atomic-write-append-in-c-or-how-do-i-get-files-opened-with-the
-            // https://stackoverflow.com/a/5469572/2055163
-            // https://stackoverflow.com/questions/1154446/is-file-append-atomic-in-unix/
+            // https://github.com/dotnet/runtime/issues/30435#issuecomment-590609103
             if ((rights & FileSystemRights.AppendData) == FileSystemRights.AppendData)
             {
                 throw new UnsupportedOperationException(
-                    String.format(
-                        "Atomic appends are not supported currently. %s",
-                        "https://github.com/dotnet/corefx/issues/39920"));
+                    "Atomically appending to files is not supported.");
             }
 
             boolean readOnly    = rights == FileSystemRights.Read;
--- a/openjdk/sun/nio/fs/NetFileSystemProvider.java
+++ b/openjdk/sun/nio/fs/NetFileSystemProvider.java
@@ -31,6 +31,7 @@ import cli.System.IO.DirectoryInfo;
 import cli.System.IO.DriveInfo;
 import cli.System.IO.File;
 import cli.System.IO.FileAttributes;
+import cli.System.IO.FileAccess;
 import cli.System.IO.FileInfo;
 import cli.System.IO.FileMode;
 import cli.System.IO.FileShare;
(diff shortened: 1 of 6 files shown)


openjdk/sun/nio/fs/NetFileSystemProvider.java
```diff
@@ -371,7 +372,40 @@ private static FileDescriptor open(String path, int mode, int rights, int share,
             if (false) throw new cli.System.IO.IOException();
             if (false) throw new cli.System.Security.SecurityException();
             if (false) throw new cli.System.UnauthorizedAccessException();
-            return FileDescriptor.fromStream(new FileStream(path, FileMode.wrap(mode), FileSystemRights.wrap(rights), FileShare.wrap(share), 8, FileOptions.wrap(options)));
+
```
DavidThi808: I think the old way should be insed an #if NETFRAMEWORK and what is here inside the #else

runtime/openjdk/java.io.cs
```diff
@@ -129,15 +129,23 @@ public static Stream open(string name, FileMode fileMode, FileAccess fileAccess)
 		{
 			return VirtualFileSystem.Open(name, fileMode, fileAccess);
 		}
-		else if (fileMode == FileMode.Append)
```
DavidThi808: Again I suggest a #if NETFRAMEWORK here, #else for the new code.

NightOwl888:
A bit confusing naming the compilation symbol `NET_CORE`. That `FileStream` constructor *does* (or at least will) exist in .NET Core 3.1. It doesn't exist in .NET Standard.

In fact, several APIs exist in .NET Core that don't exist in .NET Standard.

Suggestion: Rather than using framework targets as conditional compilation symbols, use features instead.

#if FEATURE_FILESTREAM_ATOMICITY

That way we can map features that will be compiled independently of targets.

DavidThi808:
`Suggestion: Rather than using framework targets as conditional compilation symbols, use features instead.`

I disagree. We then end up with all these various features that are tied back to the platforms. I think it's a lot cleaner to define by platform because we have 4 of those at most and so it's very clear what's what. Framework, Standard 2.0, Core 3.1, and maybe Standard 2.1.

Over the years (decades) I've found that keeping things as simple as possible is a big win. Especially as all of us aren't working closely together and haven't worked together before. It would be too easy to drop a #define one of us added that the others missed.

openjdk/sun/nio/fs/NetFileSystemProvider.java
ams-tschoening: `#if NETFRAMEWORK` doesn't work in Java and I'm somewhat sure that I changed the implementation because things failed at runtime. So, does one need to check at runtime the current version of .NET executed under and deal accordingly? Reads complicated to me:

https://stackoverflow.com/a/51974946/2055163

ams-tschoening:
> A bit confusing naming the compilation symbol NET_CORE. That FileStream constructor does (or at least will) exist in .NET Core 3.1. It doesn't exist in .NET Standard.

I marked things to be incompatible with .NET Standard and .NET Core because from my understanding even .NET Core needs an adoption, as it's not the same FileStream-CTOR provided with 3.1.

    FileInfo fileInfo = new FileInfo(path);
    fileInfo.Create(FileMode.Create, FileSystemRights.FullControl, FileShare.Read, 4096, FileOptions.None, security);

https://github.com/dotnet/runtime/issues/30435#issuecomment-590609103

ErikApption:
@timotei @mikeTWC1984 @igorkwwrd  Any updates on this? Looks like beside a review the code is ready to ship. Having even a beta version published on nuget would save us alot of time

igorkwwrd:
I don't quite understand what is going on.  Does this PR builds the .NET Core compatible IKVM?

What about this issue
https://github.com/ikvm-revived/ikvm/pull/17#pullrequestreview-413059819
where we changed the build scripts to msbuild? It doesn't build for the Core yet, but it allows to use the dotnet CLI to build the IKVM. Something like this

dotnet build ikvm.sln /p:TargetFramework=net461

And I'm working on that branch to make it build for netcoreapp3.1

ams-tschoening:
> I don't quite understand what is going on. Does this PR builds the .NET Core compatible IKVM?

Not really, this PR pre-dates all other discussed approaches like the changed build scripts. I needed to get some Java-libs be usable in some UWP-app and this PR contains changes working around compiler issues and stuff like that I came across. So things were improving compatibility with .NET Core from my point of view and @DavidThi808 suggested to bring those changes into master.

It might not be necessary at all currently and simply be kept in mind when others run into the same problems. Some things might need to be dealt with differently anyway. So this PR can be left open as a reminder or maybe even closed, because the corresponding branch [net_core_compat](https://github.com/ikvm-revived/ikvm/tree/net_core_compat) is still available or ...

ErikApption:
Thanks @igorkwwrd @ams-tschoening for the clarification. We've been looking to build a net core version and it has not been straightforward tracing all the branches. Any clues on how far ikvm is from having even an alpha on nuget that is compatible with net core? Anything in specific we can help?

ams-tschoening:
Closing this to make things easier to follow for people, while keeping the branch for now. Nobody currently seems to know if these changes are necessary at all and if so they aren't lost anyway.

PR #17, opened 2020-03-02
Port of nant build scripts to msbuild (framework)

igorkwwrd:


NightOwl888:
Great. I cloned this and started working on it. So far, I have migrated `IKVM.Reflection` and `ikvmc` to the new format (so far .NET Framework 461 is the only target).

Some of this build configuration is complete overkill moving to the new `.csproj` format. For example, every file in the `nuget` directory can now be condensed down to a few lines in a `nuget.props` MSBuild file.

The conditional strong name key is mostly done using [this method](https://www.meziantou.net/declaring-internalsvisibleto-in-the-csproj.htm) to generate the `InternalsVisibleTo` attributes. Right now, the end user is expected to pass in both the `AssemblyOriginatorKeyFile` and `PublicKey` in order to enable strong naming. I could have the build generate the `PublicKey` based on the `AssemblyOriginatorKeyFile`, but that would require checking in the small binary `sn.exe` (I believe we would need one for x86 and one for x64). Thoughts?

One thing that will clearly be an issue are all of the P-Invoke calls to Windows-only APIs. It is unclear at this point how critical they are to the main functionality, though.

igorkwwrd:
Having sn.exe checked in is okay. However I'm concerned about how much the build process will depend on the system. Would it be possible to build on Linux?

NightOwl888:
Good point. Sn.exe is not cross-platform. That said, we could make the `PublicKey` parameter required if not on Windows so the key doesn't have to be generated there, only passed in.

For now, I will leave both parameters required and leave scripting `sn.exe` until the end, and only do it if there is time to.

NightOwl888:
Well, I converted some of the projects, but when I tried converting IKVM.Runtime and IKVM.Runtime.FristPass, I started getting errors about "HideFromJavaAttribute does not exist" when the openjdk project was compiling.

Then I attempted to clear out the `bin` and `obj` folders and build from scratch, but it seems that the build has several dependencies on the DLLs from the previous build, so there is no way to start fresh. Even worse, since I can't get a build to run to create those binaries, it is in an endless cycle of build failures.

Also, it seems that since several of the project files are in the same folder, they share the same `obj` folder, and the new build process can't seem to cope with that. I think the problem is that each requires its own `project.assets.json` file to be generated and the first one in `runtime`, `DummyLibrary.csproj` generates a `project.assets.json` file that is not compatible with the others.

Another complication of the new format is that it (correctly) makes the default output folder `bin/$(Configuration)/$(TargetFramework)` rather than `bin/$(Configuration)`, so many of the targets need to be changed to account for that.

I think it is important to make the build robust enough to be able to do a "clean build" after manually deleting the `bin` and `obj` folders. But so far I haven't worked out how to accomplish this due to the fact that many of the projects require a binary to exist in order to run. Any ideas?

For now, I am thinking about starting from a new branch based on this one, and then copying over the project files one at a time, just so I can get a successful build again. But I think I should aim to move all of the project files that are in shared locations into subfolders and modify the copy logic appropriately to move the binaries back to the folders where they are now. Otherwise, the shared `obj` folder will cause issues.

igorkwwrd:
No, the build does not depend on a previous build. I just made a fresh checkout and it built without issues.

However, the projects have a particular build order involved. Ikvm.Managed.sln is set to enforce that order. If you right-click the solution in Visual Studio and open the Project Build Order item, you'll see it. It is because some projects depend on artifacts produced by the others.

The original build scripts put all public artifacts into the root bin directory. So, many msbuild projects have the AfterBuild target where they copy their output to bin/$(Configuration).

Sorry, I wasn't aware of the new format. New build system, new rules.

NightOwl888:
> Sorry, I wasn't aware of the new format. New build system, new rules.

I understand, and I didn't mean to sound like I was placing blame. I wasn't aware of this behavior, either, but then I have never tried to put more than one project in the same directory before.

After analyzing the original `.build` files, I discovered that they were outputting the build into the common `bin` directory. But several projects including `IKVM.Reflection`, `ikvmc`, and `ikvmstub` were not setup to copy their binaries there. So, I fixed that and I am making progress again.

Looking at the `openjdk.build` project and considering how the shared `obj` folder is problematic, I ended up merging them back into a single project again. I just had a successful build with that configuration. Is there any reason why it should be a separate?

> However, the projects have a particular build order involved. Ikvm.Managed.sln is set to enforce that order. If you right-click the solution in Visual Studio and open the Project Build Order item, you'll see it. It is because some projects depend on artifacts produced by the others.

Yea, I was looking at that.

It would be better if project references were set up so the "natural" build order that visual studio uses would be working. It is easier to see what depends on what. I have put some of these dependencies in. For example, rather than having a CSC call in a target build implib.exe, I created a project to build it and added a dependency on that project within `jvm.csproj`. Unfortunately, there are some circular dependencies, so this method won't work throughout the entire build.

igorkwwrd:
The projects you mentioned do not copy because their OutputDir is set to that common
bin directory.

openjdk has two parts: openjdk itself and tools and they are building at different
times.  So, two distinct projects.

Having each project in its own directory should resolve the common obj issue.  Then
some additional copying may be needed.

It's fine if you finding a better ways by splitting/combining different parts.
The current msbuild structure is not something that should be preserved we are free
to change it as needed.
I agree with the "natural" build order.

Yes, there are circular dependencies and it is why the projects like FirstPass
exist.

igorkwwrd:
@NightOwl888 's switch to the new csproj format is merged into this branch.

Added a commit https://github.com/ikvm-revived/ikvm/pull/17/commits/a7f9d3e596a8bd6342a72b5b2efa1e4a4773330e which allows targeting multiple frameworks.  Currently net461 and netstandard2.0.
So, now it's possible to select what to build on a command line, e.g.

`dotnet build /p:TargetFrameworks=net461`

Running

`dotnet build /p:TargetFrameworks=netstandard2.0`

clearly shows what is still missing on the road to the .NET Core goal.

igorkwwrd:
> FYI - There is no such thing as an executable netstandard2.0 project. Executables must target either netcoreappX.X or net4XX. Only libraries can target netstandard2.0.

Oh, right. I'll investigate. I just wanted to try a build targeting something other than net461.

NightOwl888:
No problem.

I don't believe there will be a reason to multi-target the console applications. Just targeting the latest .NET Core version should be all that we need to do. Also, it would be best to make them into "commands" of a single dotnet tool that can be deployed via NuGet. I copied the bits out of a Microsoft template to [lucene-cli](https://github.com/apache/lucenenet/tree/08e35ce8ae4ed5d47da37bf1157ad3a17928cc98/src/dotnet/tools/lucene-cli), which shows the structure of how to configure arguments, options, and commands.

That said, IMO we should wait until this is merged into master and integrated with other PRs before getting into this part.

erksch:
Hey what's the status of this? Looks pretty much approved and ready :D

tr33s:
a9989ea breaks nant building, although I'm not sure if that was intended.

CommonAssemblyInfo.cs.in was deleted in that commit, and is still referenced by ikvm.build.

CYDHJ:
use dotnet build /p:TargetFrameworks=net461 The DLL file generated by the project is successful, and the debugging report referenced to my project is as follows:
System. MethodAccessException:“[IKVM.OpenJDK.Core.dll] Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations。

Could you please help me solve this problem?

PR #18, opened 2020-03-05
New csproj Format

NightOwl888:
This branch contains everything from pull #17.

I have upgraded all of the projects to use the new `.csproj` format including the built-in generating of AssemblyInfo attributes, parameterized strong naming, and much of the setup for generating the NuGet package(s).

So far, I haven't changed any of the compilation symbols or upgraded to .NET Standard 2.0 (I did confirm that `IKVM.Reflection` will compile under .NET Standard 2.0, though), and the NuGet packaging is not complete yet.

I took a look at the original IKVM NuGet package and it contains:

- jvm.dll
- x86 and x64 native binaries
- All OpenJDK DLLs

While it is straightforward to make a NuGet package that contains these assets with this setup, the native libraries are not currently part of the solution, so cannot be built. Also, is the plan to have @jfrijters hand ownership of the original IKVM NuGet package feed over to you so we can continue releasing there?

There are still some rough edges, but I would appreciate some feedback before the work on .NET Standard 2.0 starts.


igorkwwrd:
Nice.

NightOwl888:
I chose to merge into msbuild-framework to allow these 2 PRs to be reviewed and merged together, since they are effectively both stepping stones to get us onto the new build tools to build .NET Standard libraries.

My thought was that if I chose to merge to master, I would be superseding #17 by including its changes, meaning that PR would need to be closed to merge this one. By merging it into msbuild-framework, it will effectively be a part of #17 and can be approved together by approving #17.

But, if that isn't how things should be done, I could change the target of #18 to master instead, and  #17 could be closed. Thoughts?

igorkwwrd:
I'm fine with merging #18 -> #17 -> master

Issue #22, opened 2020-06-01
Remove Old Nant Scripts


tr33s:
a9989ea breaks nant building, although I'm not sure if that was intended.

CommonAssemblyInfo.cs.in was deleted in that commit, and is still referenced by ikvm.build.

_Originally posted by @tr33s in https://github.com/ikvm-revived/ikvm/pull/17#issuecomment-636605034_

NightOwl888:
The nant scripts have been replaced with MSBuild files with the [new `.csproj` format](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj). The old nant scripts are only being kept around for reference as we work out how to finish the conversion to .NET Standard.

tr33s:
Alright, well I got the project running with MSBuild, thanks.

Requesting README and HOWTO be updated as well.

AaronRobinsonMSFT:
@NightOwl888 Is there a plan to fold some of the native build into an existing SDK project? Given the native build is so simple, it would be relatively easy to create a native build sub-task in `IKVM.Runtime.JNI.csproj`. Thoughts?

NightOwl888:
The old NAnt scripts have been removed in the main branch, and we have also updated the readme with some build details (let us know if we need to add more details). There is also now a prerelease on NuGet: https://www.nuget.org/packages/IKVM/8.2.0-prerelease0379

Issue #20, opened 2020-05-25
Building for .NET Standard 2.0


erksch:
Hey there!

Since support for .NET Standard was merged, I tried to build the dlls for .NET standard:
```
git clone git@github.com:ikvm-revived/ikvm.git
cd ikvm
dotnet build /p:TargetFrameworks=netstandard2.0
```
I got the build command from the [pull request discussion](https://github.com/ikvm-revived/ikvm/pull/17).

However I am getting the following error:
```
$ dotnet build /p:TargetFrameworks=netstandard2.0
Microsoft (R) Build Engine version 16.3.0+0f4c62fea for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 280.75 ms for C:\Users\erikz\Projects\ikvm\ikvmstub\ikvmstub.csproj.
  Restore completed in 280.73 ms for C:\Users\erikz\Projects\ikvm\jvm\jvm.csproj.
  Restore completed in 280.66 ms for C:\Users\erikz\Projects\ikvm\awt\IKVM.AWT.WinForms.FirstPass\IKVM.AWT.WinForms.FirstPass.csproj.
  Restore completed in 280.78 ms for C:\Users\erikz\Projects\ikvm\reflect\IKVM.Reflection.csproj.
  Restore completed in 280.78 ms for C:\Users\erikz\Projects\ikvm\openjdk\openjdk.csproj.
  Restore completed in 280.78 ms for C:\Users\erikz\Projects\ikvm\ikvmc\ikvmc.csproj.
  Restore completed in 280.72 ms for C:\Users\erikz\Projects\ikvm\awt\IKVM.AWT.WinForms\IKVM.AWT.WinForms.csproj.
  Restore completed in 280.78 ms for C:\Users\erikz\Projects\ikvm\ikvm\ikvm.csproj.
  Restore completed in 13.17 ms for C:\Users\erikz\Projects\ikvm\tools\SourceLicenseAnalyzer\SourceLicenseAnalyzer.csproj.
  Restore completed in 20.99 ms for C:\Users\erikz\Projects\ikvm\runtime\IKVM.Runtime.JNI\IKVM.Runtime.JNI.csproj.
  Restore completed in 39.09 ms for C:\Users\erikz\Projects\ikvm\tools\depcheck\depcheck.csproj.
[...]

PR #23, opened 2020-07-05
Support native build on *nix/Windows/macOS using SDK project

AaronRobinsonMSFT:
This can replace the use of nant for building native assets.
The native projects are now chained into the building of IKVM.Runtime.JNI
through ProjectReferences and will build both x86 and x86_64 on Windows.
On *nix and macOS only the x86_64 will be built.

The additional NuGet package - https://www.nuget.org/packages/CompilerPaths, currently only supports Win 10 SDK and VS 15+. If additional support is required for other Windows SDKs and previous versions of VS, please let me know.

Contributes to https://github.com/ikvm-revived/ikvm/issues/22

### DIFF
--- a/native/jni-native.targets
+++ b/native/jni-native.targets
@@ -1,0 +1,201 @@
+<Project>
+
+  <PropertyGroup>
+    <NativeConfiguration Condition="'$(NativePlatform)' == ''">Release</NativeConfiguration>
+    <NativePlatform Condition="'$(NativePlatform)' == ''">$(Platform)</NativePlatform>
+
+    <NativeSourceDir>$(MSBuildThisFileDirectory)</NativeSourceDir>
+    <NativeObjDir>$(MSBuildThisFileDirectory)obj/$(NativeConfiguration)/$(NativePlatform)/</NativeObjDir>
+    <NativeOutputDir>$(MSBuildThisFileDirectory)../bin/</NativeOutputDir>
+
+    <NativeName>ikvm-native</NativeName>
+    <NativeBinaryName>lib$(NativeName).so</NativeBinaryName>
+    <NativeBinaryName Condition="$([MSBuild]::IsOsPlatform(OSX))">lib$(NativeName).dylib</NativeBinaryName>
+    <NativeBinaryName Condition="$([MSBuild]::IsOsPlatform(Windows))">$(NativeName)-win32-$(NativePlatform).dll</NativeBinaryName>
+
+    <!-- Configure CompilerPaths task -->
+    <MSVCPlatform>$(NativePlatform)</MSVCPlatform>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <NativeSource Include="$(NativeSourceDir)*.c" />
+    <NativeBinary Include="$(NativeOutputDir)$(NativeBinaryName)" />
+    <Clean Include="@(NativeBinary);$(NativeObjDir)*.*" />
+  </ItemGroup>
+
+  <!--
+    Target used to build the IKVM native binary on Windows
+  -->
+  <Target
+    Name="Build-IKVM-Native-Windows"
+    AfterTargets="CoreCompile"
+    DependsOnTargets="MSVCFindCompilerPaths"
+    Inputs="@(NativeSource)"
+    Outputs="@(NativeBinary)"
+    Condition="$([MSBuild]::IsOsPlatform(Windows))">
+
+    <!-- MSVC compiler flags -->
+    <PropertyGroup>
+      <CompiledResult>$(NativeObjDir)$(NativeName).dll</CompiledResult>
+      <IncPaths>@(MSVCIncludePaths-> '/I &quot;%(rootdir)%(directory)%(filename)&quot;', ' ')</IncPaths>
+      <LibPaths>@(MSVCLibPaths-> '/LIBPATH:&quot;%(rootdir)%(directory)%(filename)&quot;', ' ')</LibPaths>
+      <SourceFiles>@(NativeSource-> '&quot;%(rootdir)%(directory)%(filename)%(extension)&quot;', ' ')</SourceFiles>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <CompilerArgs>/O2 /D WIN32 /D NDEBUG /D _WINDLL /D _MBCS /GS /W3 /nologo</CompilerArgs>
+      <LinkerArgs>/DLL user32.lib $(NativeName).res /out:&quot;$(CompiledResult)&quot;</LinkerArgs>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(NativeConfiguration)'=='Debug'">
+      <CompilerArgs>/Od /D WIN32 /D _WINDLL /D _MBCS /GS /W3 /nologo</CompilerArgs>
+      <LinkerArgs>/DLL /DEBUG user32.lib $(NativeName).res /out:&quot;$(CompiledResult)&quot;</LinkerArgs>
+    </PropertyGroup>
+
+    <MakeDir Directories="$(NativeObjDir)" />
+
+    <ItemGroup>
+      <Tokens Include="VERSIONLIST">
+        <ReplacementValue>$(Version.Replace('.',','))</ReplacementValue>
+      </Tokens>
+      <Tokens Include="VERSION">
+        <ReplacementValue>$(Version)</ReplacementValue>
+      </Tokens>
+      <Tokens Include="ARCH">
+        <ReplacementValue>$(NativePlatform)</ReplacementValue>
+      </Tokens>
+    </ItemGroup>
+
+    <!-- Generate the RC file -->
+    <TemplateFile
+        Template="$(NativeSourceDir)$(NativeName).rc.in"
+        OutputFileName="$(NativeObjDir)$(NativeName).rc"
+        Tokens="@(Tokens)" />
+
+    <!-- Generate the resource file -->
+    <Exec Command="&quot;$(MSVCWinSDKToolPath)\rc.exe&quot; /nologo &quot;$(NativeObjDir)$(NativeName).rc&quot;"
+        WorkingDirectory="$(NativeObjDir)"
+        ConsoleToMSBuild="true" />
+
+    <!-- Compile ikvm-native binary -->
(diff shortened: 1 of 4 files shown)


ams-tschoening:
@AaronRobinsonMSFT I recognized your PR already, but had the feeling that it should only be merged after referenced #22 has been closed and NANT removed entirely. Do you have another opinion?

I had additionally in mind if it would be a good idea to wait for the following PR:

https://github.com/ikvm-revived/ikvm/issues/20#issuecomment-650120019

> [...]I want to clean up the build process a bit first[...]

That reads like it might lead to conflicts?

AaronRobinsonMSFT:
> should only be merged after referenced #22 has been closed and NANT removed entirely. Do you have another opinion?

Nope that makes sense. I can more than willing to help remove NANT entirely, however this would need to be inserted prior or else we wouldn't be able to build. Perhaps I can expand this PR to include NANT removal. Would that be acceptable?

>I had additionally in mind if it would be a good idea to wait for the following PR:
>
>#20 (comment)

That is fair. Waiting for that work seems prudent. Thanks.

ams-tschoening:
> [...]however this would need to be inserted prior or else we wouldn't be able to build.

#22 was originally about NANT not working anymore anyway, so that shouldn't be a problem?

> Perhaps I can expand this PR to include NANT removal. Would that be acceptable?

Does it make anything easier for you instead of providing a second PR? With a second one we would have more control over merging this before removing NANT or afterwards and might take #20 into account etc. I think I would prefer a second PR.

native/jni-native.targets
```diff
@@ -0,0 +1,201 @@
...
+    <PropertyGroup>
+      <CompiledResult>$(NativeObjDir)$(NativeBinaryName)</CompiledResult>
+      <CompilerArgs>--shared -fPIC</CompilerArgs>
+      <SourceFiles>@(NativeSource-> '&quot;%(rootdir)%(directory)%(filename)%(extension)&quot;', ' ')</SourceFiles>
+    </PropertyGroup>
+
+    <MakeDir Directories="$(NativeObjDir)" />
+
+    <!-- Compile ikvm-native binary -->
+    <Exec Command="&quot;gcc&quot; -o &quot;$(CompiledResult)&quot; $(CompilerArgs) `pkg-config --cflags --libs gmodule-2.0` $(SourceFiles)"
```
joniles, line 115:
```suggestion
    <Exec Command="&quot;gcc&quot; -o &quot;$(CompiledResult)&quot; $(CompilerArgs) `pkg-config --cflags gmodule-2.0` $(SourceFiles) `pkg-config --libs gmodule-2.0`"
```

I found that the original command line structure worked fine on OSX, where gcc is an alias for clang/LLVM. On Ubuntu with a "real" gcc  installation, the compiler appeared to be sensitive to the order of the flags, and the resulting library would not load when compiled with the original configuration as it couldn't find the library dependencies. Ensuring that the linker flags come last seemed to fix the issue, as per the suggested change above.

NightOwl888:
@AaronRobinsonMSFT, @joniles 

I pushed some updates to sync this up with master and also spotted some missing quotes due to the fact that `[MSBuild]::IsOsPlatform()` accepts a string argument. I also changed the target frameworks to `<TargetFrameworks>net461;netcoreapp3.1</TargetFrameworks>` because the current build doesn't utilize the `PackageReference` logic to resolve those for us. However, although these updates fix the build so it finishes, these changes don't seem to be outputting the native DLLs.

Is there a way to include some stubs or something so we can build all of the native assets on a single OS? I mean, we can set up a multi-stage build so native DLLs can be built in parallel on multiple OSes in CI and then copy the pipeline assets across to the main build, but it does add some complexity that is best avoided.

I could also use some assistance to fix build issues to decouple the build from the very rigid directory structure it currently depends on. The biggest issue is that there is no way to do a solution level build because the `IKVM.OpenJDK.*.dll` files are being generated in the `openjdk` directory for both target frameworks. It would make things so much easier if we could do a solution level pack, publish, test, etc, but unfortunately a lot of stuff was copied over from nant without parameterizing it for muti-targeting compatibility or taking advantage of the `PackageReference` target framework dependency logic.

Based on the [`build.core.cmd`](https://github.com/ikvm-revived/ikvm/blob/daeb47df220208549ed22e7da07746d52fe5bacc/build.core.cmd) and [`build.framework.cmd`](https://github.com/ikvm-revived/ikvm/blob/daeb47df220208549ed22e7da07746d52fe5bacc/build.framework.cmd) files, the build order is exactly the same for each target framework, so setting up depends for that project build order should be straightforward.

I also noticed that the `IKVM.OpenJDK.*.dll` files that are being generated for .NET Core 3.1 are actually targeting .NET Framework 4.8 according to dotPeek, although I haven't tested whether that will be an issue.

wasabii:
@AaronRobinsonMSFT Can you update us with some information about your usage of the native libraries for IKVM? Are you still making use of them? If so, for what?

I have been thinking the last week about whether they're something IKVM should continue to support and would definitely want to know your use case.

AaronRobinsonMSFT:
> @AaronRobinsonMSFT Can you update us with some information about your usage of the native libraries for IKVM? Are you still making use of them? If so, for what?

I was merely trying to help here. If there is no need to build native assets then this PR is less interesting.

wasabii:
The current thinking amongst @NightOwl888 and myself is to go ahead with our 8.2 release, with .NET Core support, but without JNI support for now. We think getting a release that supports Core out is probably more pressing than having JNI working. 

And then we'll get back to this, and get it integrated into the new build layout, but probably initially without Core support, unless it somehow works out of the box. Once 8.2 is out I'm probably going to spend a week actually USING IKVM for the first time in a decade, and then I'll tackle this.

wasabii:
I have it integrated into the new build now.

So the more I state at ikvm-native, the less certain I am of it's purpose. It's got some native functions for mmap, LoadLibrary, FreeLibrary, mapped to different OS specific methods... But the heavy work is done in managed code. The JNIEnv table is declared here, but half of it is empty, and appears to be filled in by delegates generated in managed code anyways...

Everything ikvm-native does do appears to be pretty trivially done by managed code or PInvoke. We can call LoadLibrary (and do) or dlopen or dlsym..... we can call mmap... there are already managed methods for that...

Except for one thing which I'd have to look up: varargs and va_list. I seem to remember varargs being a hard thing to accomplish.

Does anybody have enough historical knowledge to know why this exists? The only thing I can come up with is for the vararg and va_list support.

wasabii:
Build system has diverged so far since this I'm going to go ahead and close it. I have a osx branch which is attempting to do the same, but it is using vcxprojs, and can xcompile to OS X from Windows. It's not quite ready either, but I think the xcompiling is something of a requirement now.

wasabii:
@AaronRobinsonMSFT I have taken a radically different approach to this. I am curious as to your thoughts.

#228

AaronRobinsonMSFT:
@wasabii Thanks for tagging me. Points for cleverness as well earned. I do like the notion of creating "mostly compiled" content to reduce set up costs, but honestly it isn't all that hard in practice.  Having a small CMake requirement would completely satisfy the need and be far less magic and entirely well understood by the community and future ikvm owners.

The current approach does have some serious limitations though that I'd like you to consider. The first is any sort of AOT scenario will be far more complex to debug and reason about. The second are platforms where allocation of executable memory is limited. This approach inherently blocks ever supporting those scenarios. Finally, the current process doesn't align with any current tooling and thus makes it fragile and "special" when looking for ikvm contributors.

Using CMake would be far simpler and could generate the assets in ways that address all the above - as a shared object (`.dll`/`.so`) or as a static (`.lib`/`.a`). That approach doesn't block anything in the future, removes any environment set up concerns and continues to work with all existing tooling.

It is an interesting approach but I'm not sure it is really getting you much other than avoiding the small/annoying integration with CMake.

wasabii:
@AaronRobinsonMSFT 

So the main problem I had when setting up the cross compiling toolsets was that it's a lot of work to set up the cross compiling toolsets. For instance, I could not manage to compile a real .dylib for OS X without the Mac OS SDK. The headers are required. And in linking the final dylib, at least dl.dylib and a few other Mac OSy things were required. And we can't distribute those. So the person working on IKVM would need to obtain them, drop them in the right location, etc. I mostly managed to get this working using 'osxcross', a cross compiling toolset put together by some guy. You give it the Mac OS SDK, and it provides a toolchain for the rest.

Except for one fun problem: osxcross only runs on Linux. And the only way to get that working today with our current CI/CD build on GitHub actions is to use WSL. And GitHub actions only has WSL1. And running LLVM on GitHub actions in WSL1 BSOD's the runner. :)

Building the Windows versions of the .dlls from Linux: same issue. Need the appropriate headers and libraries for linking.

Clang makes this a bit easier, in that it doesn't need the headers: as long as you stick to using exclusively things LLVM provides built in (va_list, etc). But you still need the libraries for linking.

The way I rationalize this is like this: What I REALLY want is for .NET to support exporting delegates with varargs and va_list support. For the .NET JIT to do the code generation at runtime. And thus obliviate the need for us to do anything native at all. But, second to that, would be using LLVM at runtime to generate the trampolines, much as I'm doing, but using OrcJIT.

But then, at the end of the day, we support a known set of platforms. It will be every platform .NET supports. A fixed list for each release. And so essentially pre-JITting what would be done at runtime isn't really a big deal. It's incredibly easy to have this code output for a new platform (I just updated the OSX branch to add osx-x64, and it took about 2 minutes).

There's something special about this ikvm-native stuff that I think makes all the difference: it shouldn't exist at all. It only exists because .NET can't export variadic function pointers. It is literally only a tiny machine code shim to jump from one calling convention to another.

If this was a different project, like, where we actually had other code in C, I'd agree 100%.

>I do like the notion of creating "mostly compiled" content to reduce set up costs

I should edit this. At the time of writing the PR, I had thought I would need the platform headers in order to generate the .ll. It turns out I don't even need that. I have reduce the .c code down so far that everything it uses is a built-in LLVM construct. It no longer even needs the headers, or the .ll files to be committed. I'm going to be changing it to do that shortly.

> The current approach does have some serious limitations though that I'd like you to consider. The first is any sort of AOT scenario will be far more complex to debug and reason about. 

Yeah. I thought long and hard about this. My view right now is that I haven't made anything worse than it was. The entirety of IKVM is based on reflection, dynamic generation of IL, etc. You can't avoid this today, nor anytime in the near future. Even a statically compiled .class file to assembly does large amounts of reflection work, and runtime code generation. The chances of this running in AOT anytime soon are minimal, and any attempt really needs to really think through what it means in Java-land to do AOT.

PR #41, opened 2022-01-25
Strong Name Key and Misc Build Features

NightOwl888:
## Updates

- Added strong name key to repo
- Added support for strong naming (not quite functioning on .NET Core because ikvmc doesn't find the key pair in the snk)
- Renamed the `build` directory to `.build` (so we can use the name later as a command)
- Migrated package reference versions to `dependencies.props`
- Moved import for `dependencies.props` to the root `Project.Build.targets` file, where we can use target framework conditions on versions
- Added OpenJDK files to .gitignore
- Added a .gitattributes file to checkout batch/bash files with the correct line endings
- Added global private reference to `Microsoft.NETFramework.ReferenceAssemblies` so the build is decoupled from the .NET Framework Developer pack
- Fixed InternalsVisibleTo attribute generation so it will correctly output the PublicKey if supplied

## Problems Found

1. Build error on strong name key for ikvmc (`build.core.cmd`)
2. `build.framework.cmd` errors complaining that we are lacking runtime identifier config settings


timmahon:
I have sent pull request #46 which should fix the strong name building for .net core.

Issue #44, opened 2022-02-13
.NET 5 vs .NET 6 breaking change?


vokounjan:
Hi, this is more of a question than an issue.

I succesfully built a .NET application that relied on a library that was available in Java only. I transpiled the library using IKVMC, then referenced the `net_core_compat` binaries and it worked like a charm targeting firstly .NET Core 3.1 and then .NET 5.

Upgrading the framework from 3.1 to 5 was seamless, required no action regarding IKVM and everything was working just fine so I believed that the framework upgrade from 5 to 6 would be no problem. However after upgrading my .csproj and Microsoft nugets, I'm receiving these runtime errors:

`TypeInitializationException: The type initializer for [name of a class] threw an exception.`

`NoClassDefFoundError: Could not initialize class java.nio.charset.StandardCharsets`

I have to say that I'm quite surprised by this. I've read some blog posts describing changes and optimizations between .NET 5 and .NET 6 but I still have no idea what could be the root cause of my problem and how to solve it.

Has anyone seen anything similar? Does anyone have any tip to try?

Thanks in advance. 🙂

NightOwl888:
Out of curiosity, does the behavior change if you [revert to NLS](https://docs.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#use-nls-instead-of-icu)?

vokounjan:
Interesting idea but unfortunately not. Changing the setting didn't make any difference but I'll probably need to read more about this topic though. Thank you for your help.

NightOwl888:
Hi there. I noticed that each of the IKVM.OpenJDK.*.dlls weren't being generated with a `TargetFrameworkAttribute`, and I suspect that might why you are getting this error. [.NET 6 is supposed to go into compatibility mode when it loads a DLL that targets .NET Core 3.1](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks), then it wasn't because we didn't specify the framework it was targeting.

Please check .NET 6 support against #52 and let us know if this fixes the problem.

NightOwl888:
I should ask, did you compile IKVM in to target .NET 6 (which we haven't tested at all), or did you compile IKVM using the .NET Core 3.1 target, transpile your assemblies in .NET Core 3.1 and then load them with a .NET 6 entry point?

vokounjan:
Hi, thank you for your response. 🙂

[...]

PR #52, opened 2022-05-11
ikvmc/CompilerClassLoader.cs: Output TargetFrameworkAttribute in IKVM.OpenJDK.*.dll

NightOwl888:
This attribute will make analysis tools such as dotPeek show the correct target framework.

It may also be the underlying cause of #44.


NightOwl888:
Superseded by #57

PR #53, opened 2022-05-21
Complete Build System / Distribution Redesign / VFS

wasabii:
I had a need for IKVM to work on .NET Core. So I started doing what I could to help it along.

I have completed a complete rewrite and restructure of the build system. Much has changed. Instead of being a direct port of the original nant project, I have structured it as a normal MSBuild project. Using multiple targets, inputs, outputs, incremental building (where possible), etc.

Except for the initial JDK8 files, you can open this in Visual Studio, and hit build, and it should succeed for both Framework and Core simultaneously. Due to limitations still present in IKVM itself, it cannot be built with the DotNet SDK. It requires Windows MSBuild. However, I have a clear path forward on this.

I have combined IKVM.OpenJDK.* into IKVM.Java. I thought long and hard about this. My view is we should respect the native distribution and layout patterns of the JDK. In the case of JDK8, everything is distributed in a single rt.jar file. However, in JDK9+, a new modular system is available. I would suggest it becomes re-split for JDK9+, in a layout that respects the Java modules as designed. For instance, the "java-base" module from JDK9 is a different thing than the IKVM.OpenJDK.Core module IKVM currently distributes. Once we get JDK9 support, we will have to map those module names into assembly names, and keeping them 1:1 will become important. So for now, I remove them, since they don't exist in JDK8. I suspect we'll end up with a IKVM.Java.Base module in JDK9. I also just really didn't see the need to include OpenJDK in the name.

The openjdk8u45-b14-stripped file that has been passed around is no longer required. I managed to complete a proper rebuild of JDK8u45-b14. This has been combined into the GitHub Actions CI/CD pipeline now in the project. The source code is checked out from OpenJDK's git repository and built on a Linux VM. This was a nightmare to untangle. JDK8u45 will only build on GCC 4.3. No modern distributions contain GCC 4.3. Old copies of GCC 4.3 won't build with more modern copies of GCC. And the switch of libglibc version complicated matters further. GitHub actions allowed me to solve this by running the JDK build step in an old copy of Debian Lenny in a docker container step nested into the build.

I have removed the *.core and *.framework source code versions. I implemented the FileSystemRights functionality in ikvm.internal.io, doing "fake-JNI-call" into IKVM.Runtime, and it is within the IKVM.Runtime code where I can now put #ifdef's based on platform. The Java classes however are identical between platforms.

The circular dependencies are still present, but managed purely within MSBuild. However, the naming conventions have changed. Projects which represent a 'reference project', containing partial implementations or only signatures, have been renamed '-ref'. I am using the features of MSBuild to make this clean: ProjectReferences are set so that the output of dependent projects is not considered a dependency. PrivateAssets, etc.

The IKVM.Java assembly generation is very different. I opted to generate the OpenJDK-side classes using a set of parallel build targets in IKVM.Java.csproj. This outputs a temporary assembly named openjdk.dll. This then gets merged into the primary IKVM.Java assembly. I picked this option so that I would not have to restructure the existing MSBuild system for C# projects. Since two assemblies are generated: IKVM.Java.dll and openjdk.dll, and then merged, the SDK-project AssemblyInfo is present in the primary assembly at the time of merge. Thus, I could remove that logic from the java-related bits. The post-CoreCompile merge step is implemented based on logic I borrowed by PostSharp.

Since MSBuild is now being used properly, we now have access to the full transitive dependency closure. That means we can use References to resolve which DLLs need to have stubs generated. No more manual listing of stubs. You simply add a ProjectReference, PackageReference or Reference, and the DLLs automatically have stubs generated. Since net461 and netcoreapp3.1 come along with different references, this is all automatically taken care of by the system on it's own. Simply add a NuGet package, and boom, it'll generate the stubs.

I noticed that ikvmstub had a -forwarders option added. I do not know the history of this, but my guess is it was done as the initial Core port. It is no longer needed. Instead, Type Forwarders are now ignored by IKVM. Which means it continues on and attempts to find the type in another assembly. Type Forwarders are not needed for BUILDING assemblies, but for referencing them at runtime from IL that is already bound to an assembly name. A situation Java does not find itself in. Instead, since all of the reference assemblies are already discovered by the MSBuild system, the types are there to be found. At runtime compilation, the real DLLs are present, there is no need for reference DLLs, however I haven't worked on this yet.

A new project, IKVM-pkg exists to generate the final NuGet package. This project is using the NoTargets SDK, so it generates no assemblies of it's own. It uses ProjectReferenes to reference the non-ref versions of the other projects, and is set to copy all of their output into it's own. PackageReferences are thus respected, and transitive NuGet dependencies appear in the final output.

Since the whole thing now builds properly, test projects are now possible. Though the DotNet SDK build tool cannot be used to build it, the DotNet SDK Test tool can be used to test it. I added a handful of test projects, but never did find any useful tests. I'd like to get some sort of minimal TCK going.

Outstanding stuff:

The biggest issue with the cross platform build are limitations in how the Core support was added to ikvmc. Instead of adding a flag to target a specific runtime version, #ifdefs were used to conditionalize certain logic based on the platform the tool was built for. Basically, ikvmc built for .NET Core cannot be used to generate proper assemblies for .NET Framework, and ikvmc built for .NET Framework cannot be used to generate assemblies for .NET Core. The expectation is that during the Core build, you'd use the Core ikvmc, and during the Framework build, you'd use the Framework ikvmc. However..... the Framework ikvmc will not run on Linux. This was sort of a misunderstanding of how toolsets are supposed to work, I believe, in the initial effort. The tool needs to behave identically on each platform, and produce output based on what you tell it to do, not where it originated from. It needs a -target argument.

The assemblies produced for .NET Core are still annotated with .NET Framework attributes. These don't seem to hurt anything, so I figure I'll leave them there. Fixing the metadata writing stuff will probably be the right time to fix that.

Next Steps:

For me, my next steps will be ensuring this actually works for what I need it for. I'd like to get some feedback and testing from others about whether this approach is better for IKVM long-term. This thing builds a NuGet package. It builds it now. Short of bugs, we could be publishing it. We should rethink versioning.

Oh. I broke stuff too. AWT is gone. I don't know if that actually matters. And the JNI native stuff. It would blow my mind if anybody was using the JNI thing.

I frankly don't know what the status of the AWT code was to begin with. Did it ever work?

My Future Plans: 

I am going to be working to clean up IKVM itself. It is old code. With old coding styles, and using old methodologies.

My first step is going to be removing the IKVM.Reflection layer. This consists of two parts, the readers and the emitters. The metadata readers seem to be a mirror of System.Reflection. I'm not sure I understand the history of this.... except that it was a long time ago. Maybe even before Cecil existed. Maybe before System.Reflection.Emit existed (in Mono?). This can be replaced by System.Reflection.Metadata pretty easily, I believe. Part of this will be removing the dual compilation. The entire code base can be using System.Type, System.Assembly, etc. There's no need for one version to use IKVM.Reflect.Type.

Second step is going to be tackling the assembly resolution and loading. A big #ifdef exists to not use ReflectionOnlyLoad on Core. Makes sense. Core doesn't support that. However, Core does support MetadataLoadContext. And so does Framework. I believe both Framework and Core should be able to support MetadataLoadContext. That will handle the resolution for the static compilation both in Core and Framework. And AssemblyLoadContext can be used at runtime for Core, with a bridge to the plain ol' AppDomain loader stuff for Framework. Cleaning this up will go a long way towards removing the #ifdefs: instead of having all these #ifdefs, just pass an AssemblyContext to the Compiler. And pass a different one for each use case.

And third, I'll probably replace the IL emitters with System.Reflection.Metadata.Ecma335.MetadataBuilder. This works on both Framework and Core. This is probably going to be mindlessly replacing certain method names with other method names.

These changes will result in a LOT less code happening in IKVM.Runtime, and I think the complete elimination of IKVM.Reflection.

We need to at least upgrade to the latest JDK8. There are probably security issues in u45-B14. I took a stab at this early on in my efforts (before discovering that I could use an old Lenny). There's enough changes to make it non-trivial, so I reverted. But yeah, the latest version of JDK8 builds fine on modern Linux. We need to escape from that crazy hack.

Tests. We should build some tests.

Anyways. That's what I've been doing the last week. What do ya'll think?


wasabii:
I'm going to tag people who might be the primary players as far as I can tell.

@NightOwl888 @igorkwwrd

wasabii:
Some changes since my first post:

I removed the IL merging step in favor of just copying the file over. A bit faster. The build process also now uses the same logic for the OpenJDK assembly as it uses for C# projects regarding AssemblyAttributes. These are generated by MSBuild (already), and saved into a temporary obj//IKVM.Java.targets.AssemblyInfo.java file which is added to the compile automatically. Basically, everything a normal build would add, gets added to the Java side.

Including TargetFrameworkAttribute. This is now handled automatically by the build. The IKVM.Java assembly is now attributed with it properly.

#55 is preventing AssemblyMetadataAttributes however, so things like RepositoryUrl, which are normally attributed, are not.

I also removed the `resources.zip` file which was attached to the assembly, and created by ikvmc using -recurse. My best efforts to understand this make me believe it was not necessary. I couldn't find anything that actually used it. Best I could tell, it was supposed to be something to allow you to more easily combine directories of jars and classes into a single assembly. And the ZIP part of it wasn't supposed to actually be in the assembly. I'm frankly not sure though. What I do know is that all the files that were in it are accessed as Java resources. And as such, work fine in resources.jar. So they're moved into resources.jar. I frankly question the wisdom of having -recurse as an option. That sort of policy is generally up to build tools.

A bug regarding IL optimization reared it's head somewhere along the lines in Core, involving the StandardCharsets. The deal was an IKVM intrinsic that was rewriting the IL for calling toCharArray into an inline char array. Incorrectly. So it literally changed it to a char[] with invalid data in it. Another thing I question the wisdom of. I strongly feel IKVM should not be doing things like rewriting the semantics behind the IL. If it's calling toCharArray, let it, and let the JIT deal with it. The job of IKVM shouldn't be to be a better Java compiler than Javac, nor a better X86/64 compiler than the CLR. So I removed it. Those classes now work fine on Core.

I rewrote the VFS layer from near-scratch. I needed to dig into the logic in some test cases to see how it was or wasn't working, but it's reliance on global state was making that a pain in the ass. Too many statics in too many places. So, rewrite from scratch. With some new test cases. The VFS now functions as a sorta-fstab/mount thing. It can have multiple mounts at multiple places. It's open to being user configured in the future. Though I suspect later user configuration might be useful to have.

The VFS is insane btw. What a thing to behold. Especially exposing java.exe into it. I'll bet money that was for Minecraft. Eh, eh? Kinda not cross platform. Will need to be addressed.

Especially `ClassJavaZipEntry`. Holy hell, that needs to die. But hasn't yet. On my list. Black magic.

NightOwl888:
> I also removed the `resources.zip` file which was attached to the assembly, and created by ikvmc using -recurse. My best efforts to understand this make me believe it was not necessary. I couldn't find anything that actually used it. Best I could tell, it was supposed to be something to allow you to more easily combine directories of jars and classes into a single assembly. And the ZIP part of it wasn't supposed to actually be in the assembly. I'm frankly not sure though. What I do know is that all the files that were in it are accessed as Java resources. And as such, work fine in resources.jar. So they're moved into resources.jar. I frankly question the wisdom of having -recurse as an option. That sort of policy is generally up to build tools.

I discovered [this post](https://sourceforge.net/p/ikvm/mailman/ikvm-developers/thread/CAGfWea7iVm3YANF98xnTPHP6-4y8wFuhMbEb8iHiMdym886vsw@mail.gmail.com/) that may provide some insight on that file.

wasabii:
Closing PR, replacing with PR from develop to master.

PR #56, opened 2022-05-24
Merge from 8.1.5717

wasabii:
It looks like back in the day on the 8.1.5717 release, Jeroen made some changes, but those changes never got merged back into the master branch. This PR isn't to actually merge those changes, but simply to tie together the 8.1.5717 branch history into master, so that version number calculations for GitVersion work properly.

It needs to be accessed whether these changes from 8.1.5717 are in fact required.


wasabii:
The commit that contained important changes, but did not appear in history, was 871a3401bd79c37372a04957328aec870a1f6460. It does appear that the changes in this commit ID found themselves into master in some other way.

NightOwl888 (approved):

Issue #65, opened 2022-05-30
System.IO.FileNotFoundException wbe


vokounjan:
`PS C:\Users\vokounjan\source\repos\Company.Project\jar\MyLibrary> C:\Users\vokounjan\source\repos\Company.Project\bin\IKVM\ikvmc.exe -target:library -out:C:\Users\vokounjan\source\repos\Company.Project\bin\MyLibrary\MyLibrary.dll -version:1.2.3 *.jar
IKVM.NET Compiler (8.2.0-prerelease.392)
Copyright © 2022 Jeroen Frijters, Windward Studios, Jerome Haltom, Shad Storhaug

*** INTERNAL COMPILER ERROR ***

PLEASE FILE A BUG REPORT FOR IKVM.NET WHEN YOU SEE THIS MESSAGE

ikvmc, Version=8.2.0.0, Culture=neutral, PublicKeyToken=13235d27fcbfff58
C:\Users\vokounjan\source\repos\Company.Project\bin\IKVM\
3.1.25 64-bit

System.IO.FileNotFoundException: Reference assemblies directory: C:\Users\vokounjan\source\repos\packs\bin.Ref\3.1.0\ref\netcoreapp3.1
   at IKVM.Reflection.Universe.BuildRefDirFrom(String runtimeDir) in D:\a\ikvm\ikvm\IKVM.Reflection\Universe.cs:line 180
   at IKVM.Reflection.Universe.get_ReferenceAssembliesDirectory() in D:\a\ikvm\ikvm\IKVM.Reflection\Universe.cs:line 156
   at IKVM.Reflection.Universe.DefaultResolver(String refname, Boolean throwOnError) in D:\a\ikvm\ikvm\IKVM.Reflection\Universe.cs:line 842
   at IKVM.Reflection.Universe.Load(String refname, Module requestingModule, Boolean throwOnError) in D:\a\ikvm\ikvm\IKVM.Reflection\Universe.cs:line 799
   at IKVM.Internal.AssemblyResolver.Init(Universe universe, Boolean nostdlib, IList`1 references, IList`1 userLibPaths) in D:\a\ikvm\ikvm\ikvmc\IKVM\Internal\AssemblyResolver.cs:line 93
   at ikvmc.IkvmcCompiler.Compile(String[] args) in D:\a\ikvm\ikvm\ikvmc\IkvmcCompiler.cs:line 178
[...]

Issue #45, opened 2022-02-14
log4j fails with The invoked member is not supported in a dynamic assembly


AliveDevil:
There is still so much more to do to support .NET Core, than just getting it build on .NET Core.
The whole dynamic-assembly modules need to be reworked in order to interop correctly with core.

I.e. Log4j crashes completely while setting up the LoggerContext:
```
Unhandled exception. System.TypeInitializationException: The type initializer for 'ch.cyberduck.core.preferences.DefaultPreferences' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'ch.cyberduck.core.preferences.Preferences' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'org.apache.logging.log4j.core.config.ConfigurationFactory' threw an exception.
 ---> System.NotSupportedException: The invoked member is not supported in a dynamic assembly.
   at Java_java_lang_Class.forName0(String name, Boolean initialize, ClassLoader loader, Class caller) in ikvm-revived\runtime\IKVM.Runtime\openjdk\java.lang.cs:line 76
   at java.lang.Class.forName0(String name, Boolean initialize, ClassLoader loader, Class caller)
   at java.lang.Class.forName(String className, CallerID )
   at org.apache.logging.log4j.util.LoaderUtil.loadClass(String className)
   at org.apache.logging.log4j.util.LoaderUtil.newInstanceOf(String className)
   at org.apache.logging.log4j.util.LoaderUtil.newCheckedInstanceOf(String className, Class clazz)
   at org.apache.logging.log4j.core.util.Loader.newCheckedInstanceOf(String className, Class clazz)
   at org.apache.logging.log4j.core.lookup.Interpolator..ctor(Map properties)
   at org.apache.logging.log4j.core.lookup.Interpolator..ctor()
   at org.apache.logging.log4j.core.config.ConfigurationFactory..ctor()
   at org.apache.logging.log4j.core.config.ConfigurationFactory.Factory..ctor()
   at org.apache.logging.log4j.core.config.ConfigurationFactory.Factory..ctor(1 x0)
   at org.apache.logging.log4j.core.config.ConfigurationFactory..cctor()
   --- End of inner exception stack trace ---
   at org.apache.logging.log4j.core.config.ConfigurationFactory.getInstance()
   at org.apache.logging.log4j.core.LoggerContext.reconfigure(URI configURI)
   at org.apache.logging.log4j.core.LoggerContext.reconfigure()
[...]

PR #67, opened 2022-05-30
Remove the magic "ReferenceAssembliesDirectory"

wasabii:
#65 describes an issue where ikvmc out of the box, on Core, isn't really able to trace down it's reference assemblies. This patch replaces the weird logic attempting to append directory names together with code that uses DependencyContext to search for reference assemblies instead. This should cause it to search in the right location.

This is an improvement over the current code, but it's still not quite right. Universe shouldn't be the place responsible for this, instead ikvmc should be adding the paths to Universe as it sees fit for it's goals, and only in the ~-nostdlib situation. But the concerns of IKVM.Runtime and ikvmc got a bit confused at some point in the past it looks like. This is at least better.


wasabii:
So, I've gone ahead and done some more stuff.

After examining how Roslyn currently works, and the effects of the -nostdlib option, and -lib and -ref, and thinking through our future usages, I think it makes sense to deprecate the -nostdlib option, as well as the default SDK search logic.

The situation is this. In the latest C# compilers (upon which ikvmc is modeled after) included with .NET Core, -nostdlib is somewhat of a no-op. This is because in Core, there is not necessarily any default SDK to use. The Core csc.exe from 6.0 could be used to compile 5.0 assemblies, or 3.1 assemblies, or even Framework assemblies. It's no longer safe for it to simply assume the user intends to compile targeting the version of the platform that csc.exe was ultimately built in.

So, with this PR, simply running ikvmc, for any platform, will result in it being unable to find any references to build against. Including mscorlib or netstandard. Instead, the user has to specify them. The user has two ways of doing this: -reference (-r) or -lib. He can either reference the assemblies specifically using -reference, or provide a search path using -lib. The simplest use case for a user targeting a known Framework, would be to use the -lib option to point to the SDK he wants to use. If he's targeting Framework, he should specify -lib:C:\Windows\Microsoft.NET\version\Framework64. If he's targeting some .NET Core SDK, he can use -lib to point to the reference assembly folder of that SDK. And if he's invoking ikvmc programmatically from within another .NET application, he has use -lib to point to his own ref/ folder (PreserveCompilationContext). It's up to him where to point. But he has to be explicit about it.

ikvmstub got the same treatment, for the same reasons.

Thus, the tools packages no longer need to distribute a compilation context.

There was a lot of very old stuff going on in ikvmc that I ended up removing as a part of this.  First, all of the stuff regarding preloading the references path based on the Framework, gone. All of the stuff about using the Framework path as a fallback, gone. There's just the single -lib option.

I stumbled onto some other legacy stuff: Apparently there used to be an old capability to compile a Stub to a DLL. This was disabled years ago, but references were sucked in from stub files to keep old systems going. We're far from that at this point, so that's removed. If you try to compile a stub, you just get a warning and it ignores the stub.

ikvmstub was also using a totally different resolution logic called LegacyResolveReference. I removed this and unified it with the logic used by ikvmc. They should be identical in this regard.

A few files got the auto format treatment while I was in them.

Code was passing around arrays to append data to. With a copy ArrayCopy methods to copy the arrays and append data. I think they did this before generics existed! They wanted strong types for the list, and ArrayList wasn't strongly typed! Replaced this with List.Add().

There is now two new test projects related to ikvmc. One uses ikvmc as a reference, and call's its main method programmatically to convert a small dynamic jar into an assembly. The other invokes ikvmc.exe as an executable to do the same. These tests don't do much except verify there is no crashing error going on. Ideally they would actually validate the output of the assembly. But not letting a release slip by with completely broken tools is a step forward.

There was a check that prevented ikvmc and ikvmstub from loading a mscorlib or netstandard that was of a higher version than the one they were built against. I do not think this was necessary. If a user wants to compile against an uplevel framework, let them. We should support our netcore3.1 tool compiling net5 assemblies.

IKVM.Reflection/Universe.cs
```diff
@@ -145,44 +149,13 @@ public enum UniverseOptions
...
-            var suffixParts = new string[] { "packs", frameworkDir + ".Ref", "3.1.0", "ref", "netcoreapp3.1" };
-            Array.Copy(suffixParts, 0, newParts, n, suffixParts.Length);
-            var dir = Path.Combine(newParts);
-            if (!Directory.Exists(dir))
-            {
-                throw new FileNotFoundException("Reference assemblies directory: " + dir);
-            }
-            return dir;
-        }
+        public static readonly string StdLibName = "netstandard";
```
NightOwl888, line 154: I thought that the `mscorlib` functionality was moved into an assembly named [`System.Private.CoreLib`](https://github.com/dotnet/runtime/blob/v5.0.0/src/libraries/System.Private.CoreLib/src/System/Int32.cs) (or at least that seems to be the case since .NET 5.0). This seems to be backed up here: https://github.com/dotnet/roslyn/issues/16211#issuecomment-270462911. Although, later they [mention](https://github.com/dotnet/roslyn/issues/16211#issuecomment-373084209) that it is 

> `System.Runtime` at design time and `System.Private.CoreLib` at runtime, but the latter is an implementation detail


Why is `netstandard` being used here? Could this be an artifact from the early attempts at targeting `netstandard2.0` and later `netstandard2.1`? And could it be the reason for #74?

I am not saying that this is wrong per se, but I am wondering if there is any documentation indicating `netstandard` is what we should be looking for.

wasabii: Because there are type forwards in it, and the code path doesn't differentiate between the static and dynamic compilers. It is not optimal.

NightOwl888: They mention [here](https://github.com/dotnet/roslyn/issues/16211#issuecomment-373085326):

> You have to pass in all assemblies that are involved in the compilation. Type forwarders are followed but you need to specify both the assembly that defines the forwarder and the assembly that the forwarder forwards to.

Is this being accounted for?

wasabii: If the user passes them in.

wasabii: That said, this PR isn't a complete overhaul of the way assemblies are located and reflected on. It's a minor fix that removes the hard coded attempt at discovering a search path, and unifies two different discovery paths. The rest of the code is still doing what it did before: expecting a core library with the core types like Object, String, etc.

It would be better, in my view, if IKVM discovered these by actually searching the assemblies.

NightOwl888: That explains why #74 is failing (and why I am getting the same error when trying to convert log4j). However, since we are not actually shipping `netstandard.dll` with IKVM tools (it isn't in the .NET Core 3.1 reference assemblies directory), what is the procedure we should give users to provide a path to it for `ikvmc`?

On my system, I have several copies, but the [advice here](https://stackoverflow.com/a/45886961) is to use `C:\Program Files\dotnet\sdk\2.0.3\Microsoft\Microsoft.NET.Build.Extensions\net461\lib\netstandard.dll`. It sounds like you need Visual Studio to be installed in order for this path to exist, though. And it is a little unsettling that it is in a `net461` folder.

It looks like there are [a couple of NuGet packages](https://github.com/dotnet/standard/issues/1565) depending on whether we need `netstandard2.0` or `netstandard2.1`. How do users discover which one to use? And what if there are dependencies that target `netstandard1.x`?

wasabii: Hmm. I had thought netstandard was around in every core sdk for compatibility.

NightOwl888: Looks like it is. But it is not in the `ref/` directory, it is under `C:\Program Files\dotnet\shared\Microsoft.NETCore.App\3.1.0`. There appears to be one for every installed SDK version. And this one is an implementation of `netstandard2.1`.

Of course, for SDKs from .NET Core 2.x, the implementation is `netstandard2.0`.

There is also a standalone reference to .NET Standard 2.1 in `C:\Program Files\dotnet\packs\NETStandard.Library.Ref\2.1.0\ref\netstandard2.1`.

Both seem to get me past the `netstandard` error.

NightOwl888: Okay, after doing it [this way](https://github.com/ikvm-revived/ikvm/issues/45#issuecomment-1141977376) (excluding `-static` and `-classloader:`), it works. It seems that the `-lib:<ikvmtools>/refs` option somehow causes it to search for `netstandard.dll`, but using `-r:<ikvmtools>/refs/*.dll` does not.

wasabii:
Implemented in develop.

IKVM.Reflection/Universe.cs
rayanngage2: > 

can I ask how to use it?

Issue #66, opened 2022-05-30
IKVM.Reflection.MissingMemberException (System.Diagnostics.DebuggableAttribute) when using IKVMC


vokounjan:
Hi,

sorry for spamming with issues but I believe that this has a different cause than #65 and it seems like I'm the first encoutering it.

After switching from previous IKVMC to current pre-release and running this command:

`PS C:\Users\vokounjan\source\repos\Company.Product> dotnet exec C:\Users\vokounjan\source\repos\Company.Product\bin\IKVM\ikvmc.dll -target:library -out:C:\Users\vokounjan\source\repos\Company.Product\bin\MyLibrary\MyLibrary.dll -version:1.2.3 -r:System.Runtime.Extensions.dll -r:System.Runtime.dll -r:System.Threading.dll -r:System.Threading.Thread.dll C:\Users\vokounjan\source\repos\Company.Product\jar\MyLibrary\MyLibrary.jar`

Now I get this problem:

```
IKVM.NET Compiler (8.2.0-prerelease.392)
Copyright © 2022 Jeroen Frijters, Windward Studios, Jerome Haltom, Shad Storhaug


*** INTERNAL COMPILER ERROR ***

PLEASE FILE A BUG REPORT FOR IKVM.NET WHEN YOU SEE THIS MESSAGE

ikvmc, Version=8.2.0.0, Culture=neutral, PublicKeyToken=13235d27fcbfff58
C:\Users\vokounjan\source\repos\Company.Product\bin\IKVM\
3.1.25 64-bit

IKVM.Reflection.MissingMemberException: Member 'System.Diagnostics.DebuggableAttribute' is a missing member and does not support the requested operation.
   at IKVM.Reflection.MissingType.__GetDeclaredMethods() in D:\a\ikvm\ikvm\IKVM.Reflection\Missing.cs:line 523
   at IKVM.Reflection.Type.GetMembers[T]() in D:\a\ikvm\ikvm\IKVM.Reflection\Type.cs:line 749
   at IKVM.Reflection.Type.GetMemberByName[T](String name, BindingFlags flags, Predicate`1 filter) in D:\a\ikvm\ikvm\IKVM.Reflection\Type.cs:line 811
   at IKVM.Reflection.Type.GetConstructorImpl(String name, BindingFlags bindingAttr, Binder binder, Type[] types, ParameterModifier[] modifiers) in D:\a\ikvm\ikvm\IKVM.Reflection\Type.cs:line 1074
[...]

Issue #64, opened 2022-05-30
Proposal: Build dotnet tool to emulate dotnet.exe (one executable with all of the commands)


NightOwl888:
ikvmc.exe was named so to model it after csc.exe, and other IKVM tools (ikvm.exe and ikvmstub.exe) are to handle other tasks, similar to the way .NET Framework used al.exe and sn.exe to handle other task. However, the new generation of .NET build tools uses a common executable ([dotnet](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet)) which contains multiple commands, each with their own parameters, options, and help.

@wasabii have been discussing this a bit, and he correctly pointed out that the tools like csc.exe and al.exe still exist, but at a lower level, and each have a singular purpose.  So this proposal isn't necessarily to eliminate the existing tools, but to roll the functionality into a [dotnet tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) that can be installed on the command line using [dotnet tool install](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-install) which will combine their functionality. 

Having a dotnet tool on NuGet is far fewer steps to deal with than downloading and extracting .zip files before running commands in build scripts and you don't need to specify a URL, just an id and a version of the package. The installer handles which binaries get installed depending on OS and platform, so although we may have separate builds for each OS they can be accessed by all users in a single way.

One option to simplify building such a tool is the [CommandLineUtils](https://github.com/natemcmaster/CommandLineUtils) package, although for [lucene-cli](https://www.nuget.org/packages/lucene-cli/4.8.0-beta00016) we ended up copying the (original) source, patching bugs, and adding support for localized resources, where the original code patched other bugs, added DI support, but is no longer accepting new features and still doesn't support localization.

Ideally, we would follow common conventions:

- Tool name: `ikvm-cli`
- Main command: `ikvm`, which other commands are nested below

Commands would be converted to match .NET

- `ikvmc` -> `ikvm build`
- `ikvm` -> `ikvm run`
- `ikvmstub` -> `ikvm stub`

We could use `ikvm` with the original parameters and simply pass them to `ikvm run` to avoid confusion in case anyone is following the old documentation.

And of course we can nest other commands and options below these, as needed.

- Parameters would be required in a specific order and are not specified with an identifier. They may be required or optional.
- Options would be changed to use `-v` (short)  and `--verbose` (long) values followed by a space ` ` rather than a colon `:` for options that accept a value.
[...]

PR #71, opened 2022-06-04
RollForward strategy for ikvmc

sergey-tihon:
@NightOwl888 I believe that with this line we will be able to run `ikvmc.exe` compiled for `netcoreapp3.1` target on machines with `net6.0` without installing `.NET 3.1` runtime. (as we do here https://github.com/sergey-tihon/OpenNLP.NET/pull/14/)

Here you can find official docs
https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward

We already use this feature in other tools for the same scenario. Example https://github.com/fsprojects/FsLexYacc/blob/master/src/FsLex/fslex.fsproj#L5-L6

### DIFF
--- a/ikvmc/ikvmc.csproj
+++ b/ikvmc/ikvmc.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp3.1</TargetFrameworks>
+    <RollForward>LatestMajor</RollForward>
     <DefineConstants>$(DefineConstants);STATIC_COMPILER;EMITTERS</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputPath>..\bin\$(Configuration)</OutputPath>


wasabii:
I don't think the tools drops are framework dependent, though.

wasabii:
I was wrong. I had intended to make them self-contained, but they were not.

So anything wrong with them being self contained?

NightOwl888:
What about the `ikvmc` build makes you think that it is framework-dependent?

According to [these docs](https://docs.microsoft.com/en-us/dotnet/core/deploying/) when excluding the RID or passing `--self-contained false`, the executable will be framework-dependent. However, [our publish process is explicitly including an RID](https://github.com/ikvm-revived/ikvm/blob/develop/IKVM.artifacts.msbuildproj#L27-L38), which I think makes it a [self-contained executable](https://docs.microsoft.com/en-us/dotnet/core/deploying/#publish-self-contained). According to [this](https://github.com/dotnet/sdk/issues/2505), `<SelfContained>` is a (apparently undocumented) property, but we are not setting it. On the other hand, the [`dotnet publish` command docs](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish) say you must explicitly include `--self-contained` or `--no-self-contained` when using RID.

But the `ikvmc.runtimeconfig.json` file is telling - it lists .NET Core 3.1 under `includedFrameworks`:

```json
{
  "runtimeOptions": {
    "tfm": "netcoreapp3.1",
    "includedFrameworks": [
      {
        "name": "Microsoft.NETCore.App",
        "version": "3.1.0"
      }
    ],
    "configProperties": {
      "System.Reflection.Metadata.MetadataUpdater.IsSupported": false
    }
  }
}
```

I don't think that the tradeoff of having a smaller deployment size is worth the hassle of having to require a .NET SDK as a prerequisite to run `ikvmc`, so self-contained seems like the right choice (at least until #64 is a thing).

And since the target framework for the output of `ikvmc` is based off of the [reference assembly location](https://github.com/ikvm-revived/ikvm/issues/66#issuecomment-1142655243), it doesn't really matter if it is running on .NET Core 3.1 since we should be able to produce `net6.0` targeted assemblies with it, right? Even if that isn't the case, having an output assembly that targets `netcoreapp3.1` should be able to run on the .NET 6.0 runtime.

BTW - it looks like there is a way to [trim the deployment size](https://stackoverflow.com/a/58678969), but we probably need to manually account for any dependencies are dynamically added through IL writing or Reflection. Not to mention, we would have to ensure all of the dependencies for all of the tools are included, since there is only 1 deployment folder for all tools.

wasabii:
And yeah, Trimming is going to be insane. We'd have to account for all of the reflection going on inside OpenJDK.

wasabii:
Well, I think something was lost in translation there. Imagine we depend on SharpZipLib. Imagine they distribute two versions of their assembly in their NuGet package, one in lib/netcoreapp3.1 and another in lib/net5.0.

We build a netcoreapp3.1 version of our project. That build ends up including netcoreapp3.1 version of SharpZipLib. User runs our code on NET5. NET5 is now executing the netcoreapp3.1 version of SharpZipLib. Will this be a problem? Probably not: but ultimately it depends on what the SharpZipLib author did.

NightOwl888:
Gotcha. Trimming definitely sounds like more trouble than it is worth, especially if these tools will later be the rare usage path.

`ikvmc` would be used both on developer machines and in CI pipelines, similar to how `dotnet build` is used (at least until we have an `ikvm build` dotnet tool). So, I guess it boils down to whether we try to minimize deployment size or number of prerequisites.

The .NET Core 3.1 End of Support is coming in December, 2022, so it seems like for the tools we should move the target to .NET 6.0 to keep up with security patches, etc. But for the class libraries, it sounds like a lot more work to deal with to upgrade and IMO, as long as `netcoreapp3.1` targeted assemblies can still be consumed, we should probably not spend the time on such an upgrade.

However, we could still support the latest security patches if we continued .NET Core 3.1 on the tools, made the deployment framework dependent, and included roll forward. With the right roll forward strategy, this would also keep them up to date when .NET 6.0 reaches end of support.

We still need to specify RID for the native bits, but as I recall there is a way that can be done with a framework-dependent deployment just by putting them in the right folders.

Since it is highly likely every dev machine and most CI environments will have either a .NET Runtime or SDK installed (or can easily happen in CI with a task), and roll forward sounds like something we need to keep the runtime up to date without changing the target from `netstandard3.1`, it seems like framework-dependent is probably the best way to go for the long-term.

wasabii:
>Doesn't minimal build system in this case mean the .NET SDK?

I mean, PROBABLY. I'm not going to fight this one.

wasabii:
>Gotcha. Trimming definitely sounds like more trouble than it is worth, especially if these tools will later be the rare usage path.

For now. One of the downsides of me merging IKVM.Java was it's size. I mean it's not terrible by today's standards, but it's not great for constrained environments.

That said, JDK9+, when we get to it, comes with a reevaluation of the breakdown of the Java libraries, into Modules, giving us an initial easy win on that front. It's not trimming, but it will help.

I do think we should replace the netcore3.1 packages with net6.0 packages. I started with netcoreapp3.1 because it was what the project was built with in mind when I came to it, and without tests, or any other sort of validation, it wasn't a problem I wanted to tackle out of the gate.

That said, everything appears to work perfectly well on .NET 6 now. So maybe that's worth reassessing for the tools drops.

NightOwl888:
> Well, I think something was lost in translation there. Imagine we depend on SharpZipLib. Imagine they distribute two versions of their assembly in their NuGet package, one in lib/netcoreapp3.1 and another in lib/net5.0.
> 
> We build a netcoreapp3.1 version of our project. That build ends up including netcoreapp3.1 version of SharpZipLib. User runs our code on NET5. NET5 is now executing the netcoreapp3.1 version of SharpZipLib. Will this be a problem? Probably not: but ultimately it depends on what the SharpZipLib author did.

This will work a little differently depending on whether the actual `netcoreapp3.1` targeted assembly is shipped or a `<PackageReference>` is used, although both of them should work.

In the former case, `netcoreapp3.1` targeted assembly should work since `netcoreapp3.1` only defines the API surface area, but if the consuming executable targets `net6.0` it will load on (at least) the .NET 6.0 runtime. But if there is a `.net6.0` targeted version of SharpZipLib, it is probably not taking advantage of all of the performance optimizations.

If `<PackageReference>` is used, then it will go through the restore dependency logic to do a "smart" selection of the TFM based on the current TFM. Only this case really "depends on what the author did" and will ship any additional API surface and conditional compilation sections that are optimized for `net6.0`. This would be the ideal way to ship the right assembly for the selected TFM (or at least re-use the logic of how it resolves package dependencies).

> I do think we should replace the netcore3.1 packages with net6.0 packages. I started with netcoreapp3.1 because it was what the project was built with in mind when I came to it, and without tests, or any other sort of validation, it wasn't a problem I wanted to tackle out of the gate.
>
> That said, everything appears to work perfectly well on .NET 6 now. Or at least, just as well as on 3.1. So maybe that's worth reassessing for the tools drops.

Alright, I was under the impression that it would involve more work. It makes more sense if this is somewhat easy to change. Of  course, we need to be careful with the above dependency resolution logic so we can still target lower TFMs than `net6.0` (since it is common for class libraries to target the lowest common API).

wasabii:
>This will work a little differently depending on whether the actual netcoreapp3.1 targeted assembly is shipped or a <PackageReference> is used, although both of them should work.

Since we're talking about the tools drop, which is a Publish, they will be shipped.

>Of course, we need to be careful with the above dependency resolution logic so we can still target lower TFMs than net6.0 

Still only talking about the tools drop.

NightOwl888:
> > This will work a little differently depending on whether the actual netcoreapp3.1 targeted assembly is shipped or a  is used, although both of them should work.
> 
> Since we're talking about the tools drop, which is a Publish, they will be shipped.
> 
> > Of course, we need to be careful with the above dependency resolution logic so we can still target lower TFMs than net6.0
> 
> Still only talking about the tools drop.

Gotcha. The [dependency resolution from NuGet is documented here](https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution). The `<PackageReference>` dependency resolution logic is codified in [NuGet.DependencyResolver.Core](https://www.nuget.org/packages/NuGet.DependencyResolver.Core/). This seems like the best long-term option, but probably doesn't belong in `ikvmc` if the plan is to keep it dumb and not know anything about TFMs.

Failing that option, it sounds like the best case scenario is to ship the lowest common TFM in the drop `netcoreapp3.1` or higher. I don't think we need to go lower than that.

wasabii:
Sorry, I am confused. Are you talking about the tools package or something else?

NightOwl888:
> Sorry, I am confused. Are you talking about the tools drop or something else?

Both.

You were asking about how the drop was supposed to figure out which TFM to use for platform libraries and 3rd party packages (such as SharpZipLib). Ideally for the long term we would use [NuGet.DependencyResolver.Core](https://www.nuget.org/packages/NuGet.DependencyResolver.Core/) to determine which assembly to include, then run `ikvmc` and pass the resolved assembly as dependency ([specified with `-r`](https://github.com/ikvm-revived/ikvm/blob/develop/ikvmc/IkvmcCompiler.cs#L319)).

But until we have a higher-level tool that calls `ikvmc`, it seems that it would be best to simply deploy the `netcoreapp3.1` assemblies (even if the tools actually target `net6.0`) and include them by default when running `ikvmc`.

Since you were talking about removing the `-nostdlib` option, perhaps for the time being we could use it as a signal. If it is specified (or defaulted) use the dependencies from the drop. If not, they would need to be [specified with `-r`](https://github.com/ikvm-revived/ikvm/blob/develop/ikvmc/IkvmcCompiler.cs#L319). Then when the `-nostdlib` option is removed, specifying these dependent assemblies on the command line will be a requirement and we would no longer include a specific TFM of them in the drop. But at that point we should have a higher-level tool to handle those details about TFM. Does that sound reasonable?

wasabii:
>You were asking about how the drop was supposed to figure out which TFM to use for platform libraries and 3rd party packages (such as SharpZipLib).

It doesn't figure it out. The TFM you publish it as is what it uses. This isn't our choice. This is how it works. The SharpZipLib.dll file sits in a folder next to ikvmc.exe, and that's the end of it.

Higher level tools can't fix this. That's how .NET works.

Nothing about -stdlib effects how SharpZipLib is located in my example. You are confusing the published executable vs what the executable does in converting DLL files. These are not related.

NightOwl888:
> > You were asking about how the drop was supposed to figure out which TFM to use for platform libraries and 3rd party packages (such as SharpZipLib).
> 
> It doesn't figure it out. The TFM you publish it as is what it uses. This isn't our choice. This is how it works. The SharpZipLib.dll file sits in a folder next to ikvmc.exe, and that's the end of it.
> 
> Higher level tools can't fix this. That's how .NET works.
> 
> Nothing about -stdlib effects how SharpZipLib is located in my example. You are confusing the published executable vs what the executable does in converting DLL files. These are not related.
> 
> > Does that sound reasonable?
> 
> It sounds confused. Like, I really think you are not correctly understanding what the tools drop is, and what the TFM for that drop is about.

Ahh, okay. I was confused. I thought that the `ikvmc` dependencies were somehow inherited by the output assemblies. I guess they are when you consider that the IKVM NuGet package depends on them, but in that case the TFM is already resolving correctly.

wasabii:
Yeah, like you're talking about stuff inheriting TFMs.... but assemblies don't bind by TFMs. They bind by AssemblyName, and on Framework including Version and PublicKey with the option for runtime redirects.

wasabii:
So I'm going to close this one for now. We don't distribute non-NetCore3.1 framework-independent versions of the tools. And, in my current experience, running the tools on a later environment doesn't quite work right yet. They're going to need some work before I feel comfortable saying they're supported on later runtimes.

sergey-tihon:
> And, in my current experience, running the tools on a later environment doesn't quite work right yet. They're going to need some work before I feel comfortable saying they're supported on later runtimes.

It is sad. @wasabii can you please leave a little bit more details about issues you faced `running the tools on a later environment`?

PR #75, opened 2022-06-07
Added docs from SourceForge WIKI

NightOwl888:
This cleans up the WIKI docs, fixes broken links, and much of the grammar. And most importantly, brings the official docs into our repository so we can edit them and keep them in sync with the source code.

The docs for the tools are pretty solid at this point - I scoured the source code to make sure all of the command line options are in the documentation. Although, the usage examples for the tools probably need to be updated.

Most of the docs in the top level `docs/` directory still need to be updated for the .NET Core bits, which I could use some help with. There are several references to .NET Framework, GAC, Mono, renaming IKVM executables to Java executables, etc. that still need to be fixed. The biggest issues are in the following files:

- [ ] `debugging.md`
- [ ] `installation.md`
- [ ] `tutorial.md`
- [ ] `user-guide.md`

Plus we are missing the following from the distribution (required for the tutorial).

- [ ] `hello.jar`

I think we can probably condense the above into a single tutorial document and perhaps change the `installation.md` into a `tools/installation.md` document, since using the libraries is as simple now as using the NuGet package.

Oddly, the `debugging.md` instructions use the `vfs.zip` in order to enable Visual Studio debugging. However, we will probably need to comment out the whole section until we have support for portable `.pdb` format.

The following info is also missing from the current documentation:

- [ ] Example demonstrating how to choose target framework/reference assemblies for `ikvmc` on .NET Framework
- [ ] Example demonstrating how to choose target framework/reference assemblies for `ikvmc` on .NET Core
- [ ] Example demonstrating how to add a `TargetFramworkAttribute` to the output assembly of `ikvmc`


wasabii:
@NightOwl888 Since this is just a bunch of docs, I'm going to pull them into develop, and put them in the right place for the new site. Unknown on their quality, but we can get to that later.

Issue #79, opened 2022-06-10
Jar file worked fine using ikvm and fail to work when compiled by ikvmc


xjjius:
I was trying to use a pure java cheminformatics tool named OPSIN (https://github.com/dan2097/opsin) in my NET project. The opsin tool provides both a cli tool and share libs for programing. 

The cli tool of opsin worked perfect with ikvm (for NET 6): 

`ikvm -jar opsin-cli-2.6.0-jar-with-dependencies.jar`

Since it acted the same as using local jvm ( java -jar ...) except a minor warning like:
`Unable to get Charset 'cp65001' for property 'sun.stderr.encoding', using default UTF-8 and continuing.`
I assumed that the ikvm runtime has no big problem running this tool. 

However, when I was trying to compile the jar file to exe file, the result exe is not working. I did get the jar file compiled with the `-nostdlib -r:<ikvmtools>/refs/*.dll` workaround. Here is the warning message during the compiling process:

`ikvmc -nostdlib -r:./refs/*.dll D:\Download\Cheminformatics\opsin-cli-2.6.0-jar-with-dependencies.jar`
IKVM.NET Compiler (8.2.0-prerelease.809)
Copyright © 2022 Jeroen Frijters, Windward Studios, Jerome Haltom, Shad Storhaug

note IKVMC0004: Using main class "uk.ac.cam.ch.wwmm.opsin.Cli" based on jar manifest
warning IKVMC0127: Assembly ".\refs\netstandard.dll" is ignored as previously loaded assembly "D:\test\ikvm\netstandard.dll" has the same identity "netstandard, Version=2.1.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51"
note IKVMC0002: Output file is "opsin-cli-2.6.0-jar-with-dependencies.exe"
warning IKVMC0100: Class "org.osgi.framework.BundleActivator" not found
warning IKVMC0100: Class "com.lmax.disruptor.EventFactory" not found
warning IKVMC0100: Class "com.fasterxml.jackson.dataformat.xml.JacksonXmlModule" not found
warning IKVMC0100: Class "com.fasterxml.jackson.databind.ser.std.StdScalarSerializer" not found
warning IKVMC0100: Class "com.lmax.disruptor.ExceptionHandler" not found
warning IKVMC0100: Class "com.fasterxml.jackson.core.type.TypeReference" not found
[...]

PR #81, opened 2022-06-10
Load non-default codepages for better compatibility with Java on .NET Core

NightOwl888:
See [#79#issuecomment-1152260713](https://github.com/ikvm-revived/ikvm/issues/79#issuecomment-1152260713)

### DIFF
--- a/IKVM.sln
+++ b/IKVM.sln
@@ -57,11 +57,13 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ikvm-tests-native-linux", "
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ikvm-tests-native", "ikvm-tests-native", "{B6083619-C4E8-4D8E-95BD-217E3663134C}"
 EndProject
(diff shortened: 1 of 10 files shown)


wasabii:
I'm not totally understanding how this relates to the JDK.

NightOwl888:
> I'm not totally understanding how this relates to the JDK. Loading code pages in .NET has nothing to do with what encodings are available in the JDK.

You have a good point. We probably don't need the encoding providers.

However, I ran into a snag when trying to add tests for the `ikvm` console app since we already have a project named `IKVM.Tests`. I suggest we move the console apps into a `src/tools` subfolder so we can keep the naming consistent, but not run into collisions between similarly named class library projects. Happy to close this PR and do that separately, but I just wanted to check with you whether moving them would cause build issues.

wasabii:
My current thinking has been to rename them, not move them. Once the namespaces and stuff are figured out. To keep them consistent with that.

Like, IKVM.Exe, IKVM.Compiler.Exe, IKVM.StubGenerator.Exe, or something.

I just haven't got deep enough into each of the components to actually know what the namespaces should be.

PR #84, opened 2022-06-27
ci: publish github releases

viceice:
Currently the github releases will be generated as draft, so the nuget packages and the sdk are out of sync


wasabii:
This was kind of on purpose, to give a chance to fix up the release notes, etc.

viceice:
can also easily fixed after 🤷‍♂️

currently all never releases are missing which is more worth that wrong release notes

wasabii:
Yeah. Maybe you're right. Can always fix the notes after anyways.

viceice:
@wasabii Can you trigger a new release, so we get matching nuget and github releases ?

viceice:
or just publish the drafts?

PR #103, opened 2022-07-11
java.net.Sockets rewrite

wasabii:
Complete rebuild of the java.net.Socket and java.net.DatagramSocket stack. No longer copying the OpenJDK Windows implementation with modifications. Instead, using the Solaris Java classes, with a fairly simple mJNI backend in C#. So much cleaner.

Should get sockets working on Linux.


wasabii:
Looks like I'm going to have to redo Net.cs (nio) to finish this though.

PR #109, opened 2022-07-14
Remove SaveDebugImage

wasabii:
Feature is impossible to reimplement on Core given the current state of things. I don't think we should have big features like this which have no future path forwards.

It might make a return some day when we replace System.Reflection.Emit. But that's going to be far down the road, and whatever end up is going to be very different.

### DIFF
--- a/src/IKVM.Runtime/DynamicClassLoader.cs
+++ b/src/IKVM.Runtime/DynamicClassLoader.cs
@@ -115,20 +115,6 @@ namespace IKVM.Internal
             this.moduleBuilder = moduleBuilder;
             this.hasInternalAccess = hasInternalAccess;
 
-#if !STATIC_COMPILER
-            if (JVM.IsSaveDebugImage)
-            {
-                if (saveClassLoaders == null)
-                {
-                    System.Threading.Interlocked.CompareExchange(ref saveClassLoaders, new List<DynamicClassLoader>(), null);
-                }
-                lock (saveClassLoaders)
-                {
-                    saveClassLoaders.Add(this);
-                }
-            }
-#endif
-
 #if STATIC_COMPILER || CLASSGC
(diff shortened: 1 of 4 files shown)


AaronRobinsonMSFT:
> It might make a return some day when we replace System.Reflection.Emit. But that's going to be far down the road, and whatever end up is going to be very different.

We are currently prototyping this work - see https://github.com/dotnet/runtimelab/tree/feature/assembly-builder-save. We are working through some ideas for only Save in memory, not Run _and_ Save - which was possible in NetFX. The plan would be to make Run and Save the default. I'd say the earliest would be post-.NET 8.

wasabii:
@AaronRobinsonMSFT 

Ahh. Nice. That's good to hear. It was always such a weird thing to see removed, and I've had to work around it in a number of other projects.

Do you have any idea on what the timeline might be? Sounds a couple years out. Might effect my ideas of migrating the assembly generation to System.Reflection.Metadata.

wasabii:
Haha. You're experiment is using Metadata.Ecma under the hood anyways.

That was generally my idea. Replace the core of the assembly generation with Metadata.Ecma335, use it to generate assemblies for everything, including dynamic runtime stuff. I've been thinking about it a lot lately, and wondering if it would be feasible to pump an assembly generated with Metadata.Ecma335 into an AssemblyBuilder, so as to get unloading on Framework.

AaronRobinsonMSFT:
> Do you have any idea on what the timeline might be? Sounds a couple years out. 

If the prototype is successful I could see this being in .NET 8 (i.e., 2023-NOV), however that depends on funding and a fair bit of star alignment. Since we haven't even started .NET 8 planning yet I'm not sure. I can say there are several stakeholders for the feature so momentum is there, it is weighting it against the other needs that is tricky. It would be a rather large feature so likely announced widely, I'd monitor the normal channels for updates/announcements.

Issue #118, opened 2022-08-01
IKVM.NET.Sdk


wasabii:
Proposal for a new `IKVM.NET.Sdk` package that allows one to build Java projects, with .java source code.

Users Would add a `<Project Sdk="IKVM.NET.Sdk" >...</>` tag onto their `msbuildproj` files to opt in.

Sdk would be modeled after Microsoft.NET.Sdk, and likely inherit from it. Support multiple language, Java being the primary one (Kotlin anybody, woh!)

Would use standard MSBuild ItemGroups: `Compile` would point to .java files.

Compiler could be implemented in IKVM itself (javax.tools).

SDK would be used to replace the build logic in IKVM.Java itself in a bootstrap mode.

Eventual Language Service. Could make use of ECJ.

Eventual VSIX (or whatever they're calling them now) with a .ikvmproj extension or some such.

wasabii:
I'm working on this now in order to implement IKVM.JavaTest in the project, as that project has required me to actually import the JTReg source code and make changes.

GeorgeS2019:
[This: KVM.NET.Sdk](https://github.com/ikvm-revived/ikvm/pull/122) is super exciting

wasabii:
This issue is the discussion. That nobody else participates is unfortunate.

GeorgeS2019:
@wasabii  

Have you considered updating[ IKVM wiki?](https://en.wikipedia.org/wiki/IKVM.NET)  .NET6 is disruptive after ~9 years

wasabii:
IKVM.NET.Sdk has been mostly finalized. Hopefully I can get it in today.

One thing to note is that it will only support TFMs of net461 and netcoreapp3.1, with the former only buildable on Windows. The first part of this should be fine, since building netcoreapp3.1 projects doesn't preclude them from being consumed in higher platforms. The former is a continuation of the requirement that only the Framework tools build Framework assemblies. And Framework tools won't run on Linux.

GeorgeS2019:
Incredible!

NightOwl888:
> Framework tools won't run on Linux.

@wasabii - I have several projects setup to build .NET Framework DLLs on Linux, and I have confirmed it works. The .NET SDK builds .NET Framework on Linux, but there is no way to run the tests on Linux.

[...]

PR #122, opened 2022-08-14
IKVM.NET.Sdk

wasabii:
Split the IKVM.Java build logic mostly out of IKVM.Java into IKVM.NET.Sdk, a new SDK package. Depends on Microsoft.NET.Sdk. But supports full builds of source .java files all the way to an assembly.

Currently doesn't work stand alone. But should be good enough to build IKVM.Java.

closes: #118


wasabii:
Template logic got replaced with T4.

wasabii:
VERSION file in package replaced by targets/generated.props which gets included.
Split the build tasks out into a separate NuGet package: IKVM.MSBuild. This reduces the size of the main IKVM package itself. However, that package does have a dependency to IKVM.MSBuild for now for compatibility. Might make sense to split IkvmReference stuff out. But not doing that yet.
Added a IkvmImporter task. Might need to rename to IkvmExporter? Basically a task for ikvmstub.

Issue #137, opened 2022-08-30
java.beans.Introspector.findExplicitBeanInfo(Class beanClass) may throw unhandled ClassNotFoundException in .cctor


oising:
This popped up this morning in org.apache.log4j.LoggerManager's c# static cctor as:

```
Unhandled exception. System.TypeInitializationException: The type initializer for 'org.semanticweb.elk.cli.Main' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'org.apache.log4j.LogManager' threw an exception.
 ---> ClassNotFoundException: org.apache.log4j.PatternLayoutBeanInfo
   at IKVM.Java.Externs.java.lang.Class.forName0(String name, Boolean initialize, ClassLoader loader, Class caller)
   at java.lang.Class.forName0(String name, Boolean initialize, ClassLoader loader, Class caller)
   at java.lang.Class.forName(String name, Boolean initialize, ClassLoader loader, CallerID )
   at com.sun.beans.finder.ClassFinder.findClass(String name, ClassLoader loader)
   at com.sun.beans.finder.InstanceFinder.instantiate(Class type, String name)
   at com.sun.beans.finder.InstanceFinder.find(Class type)
   at java.beans.Introspector.findExplicitBeanInfo(Class beanClass)
   at java.beans.Introspector..ctor(Class beanClass, Class stopClass, Int32 flags)
   at java.beans.Introspector.getBeanInfo(Class beanClass)
   at org.apache.log4j.config.PropertySetter.introspect()
   at org.apache.log4j.config.PropertySetter.getPropertyDescriptor(String name)
   at org.apache.log4j.config.PropertySetter.setProperty(String name, String value)
   at org.apache.log4j.config.PropertySetter.setProperties(Properties properties, String prefix)
   at org.apache.log4j.config.PropertySetter.setProperties(Object obj, Properties properties, String prefix)
   at org.apache.log4j.PropertyConfigurator.parseAppender(Properties props, String appenderName)
   at org.apache.log4j.PropertyConfigurator.parseCategory(Properties props, Logger logger, String optionKey, String loggerName, String value)
   at org.apache.log4j.PropertyConfigurator.configureRootCategory(Properties props, LoggerRepository hierarchy)
   at org.apache.log4j.PropertyConfigurator.doConfigure(Properties properties, LoggerRepository hierarchy)
   at org.apache.log4j.PropertyConfigurator.doConfigure(URL configURL, LoggerRepository hierarchy)
[...]

PR #146, opened 2022-09-11
More SDK changes

wasabii:
Moved around the build targets quite a bit and improved build speed.
More advanced IkvmReference support in in-tree build.
IKVM.Java.Extensions project introduced. Currently not exposed out. But eventually should be. Contains a number of collection wrapper classes currently. I suspect this should be a separate NuGet package eventually.
Added new -jni argument to ikvmc. Allows specification of the Runtime.JNI assembly specifically, to override the default detection of it being in the same directory of Runtime.dll. During in-tree builds, using ProjectReferences, the DLL ends up being located in it's own project dir.


wasabii:
+fixes #137

PR #170, opened 2022-09-26
Java Runtime Image

wasabii:
This PR begins making use of the new runtime image. The VFS has been altered to not overlap or pretend to be a java.home, and retains only two pieces of functionality: cacerts generation, and the assembly directory. It is now rooted in \\.ikvm and /mnt/.ikvm.

VMSystemProperties has been cleaned up to detect ikvm.properties and calculate the appropriate java.home from that value.

The VFS has become much smaller.

Build output directory has been altered to 'dist' since that makes a bit more sense.


wasabii:
In an effort to reduce build times, IKVM.Image.JDK-bin and JRE-bin are used to pass through the combined contents of each individual tool project. Given MSBuild properly maintaining state,, this should reduce the number of passes through the various utility code.

The ikvmstub production still seems to be the slowest single part though. And probably the easiest to fix. Will fix outside this PR though.

Issue #184, opened 2022-10-17
IkvmReference does not display warnings for cached entries


wasabii:
IkvmReference cache is necessary for ensuring that generated assemblies don't regenerate when not required, given the time and complexity of doing them. However, if they're already cached, it's a bit hard to discover what went wrong into generating them in the first place. Since you have to catch the first build, which may be design time.

The output of the ikvmc command should be captured, added to .log next to the dll, at least.

PR #190, opened 2022-10-25
Copy response and log files for compiler to cache directory

wasabii:
Will allow the user to review what went into generating a cached assembly.


wasabii:
fixes #184

Issue #198, opened 2022-10-29
Windows ARM support (Runtime)


wasabii:
SDK support will be absent since Windows ARM32 is largely abandoned, except for IoT devices, where it makes sense to have runtime support.

paulirwin:
Hi, thanks for this! I am getting an error trying to build a project that was previously using 8.1 but now has been updated to use 8.4.5 and .NET 6. Attempting to build with Visual Studio 2022 17.5 on Windows on ARM. Getting the following error:

```
{truncated_path}\.nuget\packages\ikvm.msbuild\8.4.5\buildTransitive\netstandard2.0\IKVM.MSBuild.Tasks.targets(25,9): error MSB4044: The "IkvmGetReferenceAssemblies" task was not given a value for the required parameter "ToolPath".
```

I have the latest Microsoft Open JDK build for arm64 installed. Any ideas on what to try?

wasabii:
Um. I'm a bit confused what the question is about.

This bug was about runtime support for ARM (not ARM64).

You're asking about building using IKVM. Which would be SDK support. Which I believe is #197. Which is currently unsupported. And I'm pretty sure it never has been.

paulirwin:
@wasabii Thanks, I did not realize there was that distinction. I'll follow the other issue for updates.

PR #201, opened 2022-10-30
Windows ARM Support (Runtime)

wasabii:
Basic generation of Windows ARM binaries, packages, etc. Unsure where we are on stability.


wasabii:
closes #198

PR #202, opened 2022-11-01
Fixes invalid codepage (1) error in JNI

AliveDevil:
Multiple arguments passed to JNI are encoded in modified UTF-8, which StringFromOEM violates.

> name: a fully-qualified class name (that is, a package name, delimited by “/”, followed by the class name). If the name begins with “[“ (the array signature character), it returns an array class. The string is encoded in modified UTF-8.

https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html

OEMCodePage can return an invalid codepage (1) on non-standard cultures (en-de, en-ch)


src/IKVM.Runtime.JNI/JNIEnv.cs
```diff
@@ -349,8 +349,7 @@ internal static string StringFromOEM(byte* psz)
             {
                 if (psz[i] == 0)
                 {
-                    int oem = System.Globalization.CultureInfo.CurrentCulture.TextInfo.OEMCodePage;
-                    return new String((sbyte*)psz, 0, i, Encoding.GetEncoding(oem));
+                    return new String((sbyte*)psz, 0, i, Encoding.Default);
```
AliveDevil, line 352: I changed this here to Encoding.Default, because JNI[^1] uses this `StringFromOEM` for creating the JVM Args.
[^1]: https://github.com/ikvm-revived/ikvm/blob/0f894be58dcc287ff996a87cfa64ade5917f0780/src/IKVM.Runtime.JNI/JNI.cs#L219

Didn't want to create any issues with Sun UTF-8 encoding here, as that might introduce other issues.

wasabii: Encoding.Default is definitely wrong too. The spec says modified UTF-8. There's a function for that.

wasabii:
So here's what I've discovered about JavaVMOptions:

The specifications declare it as "platform default encoding". They do not declare it as UTF. They do not declare it as ASCII. But "platform default".

On .NET Framework Encoding.Default is correct. This value, on Framework, is derived from the System Default Encoding for Windows. This can change depending on the configuration of the host OS.

.NET Core declares that Encoding.Default will always return UTF8. Regardless of the platform it is running on.

Encoding.Default is the correct answer, for Framework. But it is no the correct answer for Core.

So, reading trough the hotspot code. The JavaVMOptions are parsed using String.makePlatformDefault(). makePlatformDefault derives it's encoding class from a System property (sun.jnu.encoding). This system property is set by hotspot, for Windows, to the result of GetSystemDefaultLCID: https://github.com/openjdk/jdk8u/blob/f04ad96cf53385c9f8aa071a4167ad7790cb8466/jdk/src/windows/native/java/lang/java_props_md.c#L705

However, if the user default encoding is 0c04 and the OS is Vista, it overrides that to MS950_HKSCS. We don't support Vista, so we don't need to worry about that.

GetSystemDefaultLCID is defined here: https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getsystemdefaultlcid

IKVM sets "sun.jnu.encoding" incorrectly as well, in VMSystemProperties.java. In IKVM we set it to Encoding.Default. This means our sun.jni.encoding property, on Core, would return UTF8, while OpenJDK would instead be using whatever the local system encoding is. This would however be CORRECT on Framework, where Encoding.Default is equivalent to GetSystemDefaultLCID. So this is probably another bug in IKVM on Core that we should fix.

So, the CORRECT ANSWER for parsing JavaVMOptions, is to use Encoding.Default on Framework, but get the appropriate System encoding on Core.

It looks like obtaining the correct value for both can be done using:

```
System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance);
var enc = System.Text.Encoding.GetEncoding(0);
```

PR #225, opened 2022-11-25
Enable CLASSGC

wasabii:


wasabii:
Deleting this guy. CLASSGC seems pretty broken. Will need to reevaluate. Probably is going to consist of a number of individual changes over time that stop holding hard references, so things are collectable by default. Then, some per-ClassLoader mode which determines whether the assembly is unloadable. There are ramifications to that. No P/Invoke, few other things, so it sort of matters.

PR #226, opened 2022-11-26
Fix references to IKVM.refs.props

Beyley:
The fact the case didnt match the actual files caused problems when building under case-sensitive filesystem, as is customary on Linux


wasabii:
Ahh.

There are other problems with building on Linux, though. Namely that the entire net461 side can't build.

Beyley:
> Ahh.
> 
> There are other problems with building on Linux, though. Namely that the entire net461 side can't build. Nor can the vcxproj's work.

The net7 sdk on linux can actually fully build net461 apps, and you can run them through Mono

wasabii:
>The net7 sdk on linux can actually fully build net461 apps, and you can run them through Mono, but yes the vcxproj are also a problem

Yeah, this is an approach I just started fooling with today.

One of our biggest issues is that the static compiler can only produce code for the TFM it was built with. Too many references to runtime types.

Hence, to produce net4 assemblies, you must run the net4 version of ikvmc. To produce stubs from net4 assemblies, you must run the net4 version of ikvmstub.

Beyley:
Im not sure what the CI breakage is, it may just be classic github actions being flakey, im going to merge develop into my branch and see if CI is happy

Beyley:
i am terrible with git i tried to retroactively sign the commit and i messed everything up bruh

Beyley:
k fixed

PR #235, opened 2022-12-16
Feature/unsafe

wasabii:
Rebase Unsafe on the OpenJDK implementation. Move all of the changes into Unsafe.cs (managd JNI).

Remove the intrinsic usage. This seems of limited usage. Very hard core performance optimizations, that we should reestablish justification for.


wasabii:
Part of this change is removing the field ID table from Unsafe, and replacing it with usage of Cookie from FieldWrapper. Cookie is an IntPtr, ultimately a MemberWrapper, and a GCHandle. It is native sized. So, it'll probably not work on the older 32bit versions of the Unsafe methods. It would be nice to have some sort of integer based Field ID.... but maintaining a local table of them doesn't seem like a decent way to go about it.

wasabii:
Switch from using Cookie from FieldWrapper. Instead, have a FieldDelegateRef, which maintains a GCHandle to itself. We can give out the IntPtr of the GCHandle, and do a lookup to the delegate very quickly. Pointer to instance to instance. Better than scanning a table.

Cleaned up the FastMethodAccessorImpl. Started this because of a bug where a main() method in a test was being removed from a StackTrace. Turns out to be the old x64 Framework JIT issue with it doing tail calls. Something about our Unsafe changes changed enough that the JIT now decided to do a tailcall on a main entry point in the test. The old IKVM hack to avoid tail call optimizations wasn't being enabled in the case of a discarded tail call that returns (separated from invoke with pop). Fixed that.

Issue #80, opened 2022-06-10
NullPointerException when running Maven with ikvm command


NightOwl888:
## Command Output / Stack Trace

<details>
  <summary>Click to expand!</summary>

```console
---------------------------------------------------
constituent[0]: file:/F:/mvn-sample/apache-maven-3.8.5/conf/logging/
constituent[1]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/commons-cli-1.4.jar
constituent[2]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/commons-io-2.6.jar
constituent[3]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/commons-lang3-3.8.1.jar
constituent[4]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/guava-25.1-android.jar
constituent[5]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/guice-4.2.2-no_aop.jar
constituent[6]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/jansi-2.4.0.jar
constituent[7]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/javax.annotation-api-1.2.jar
constituent[8]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/javax.inject-1.jar
constituent[9]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/jcl-over-slf4j-1.7.32.jar
constituent[10]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-artifact-3.8.5.jar
constituent[11]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-builder-support-3.8.5.jar
constituent[12]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-compat-3.8.5.jar
constituent[13]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-core-3.8.5.jar
constituent[14]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-embedder-3.8.5.jar
constituent[15]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-model-3.8.5.jar
constituent[16]: file:/F:/mvn-sample/apache-maven-3.8.5/lib/maven-model-builder-3.8.5.jar
[...]

PR #241, opened 2022-12-27
Fix classloader

DjArt:
There fixes for class loader, including, probably https://github.com/ikvm-revived/ikvm/issues/80. It's a first step for fixing WinForms AWT.


wasabii:
Can you give an explanation of some of these changes, and the theory behind them?

DjArt:
@wasabii yes, but firstly I want to fix conflicts, so little bit later

src/IKVM.Runtime/TypeWrapper.cs
```diff
@@ -3605,6 +3605,11 @@ internal virtual byte[] GetFieldRawTypeAnnotations(FieldWrapper fw)
             return null;
         }
 
+        public bool Equals(TypeWrapper other)
```
wasabii, line 3608: This can't be right. The name of a type wrapper isn't guaranteed to be unique, as the name of types themselves aren't guaranteed to be unique.

src/IKVM.Runtime/Java/Externs/java/lang/Class.cs
```diff
@@ -46,7 +47,12 @@ static class Class
                 global::java.lang.SecurityManager sm = global::java.lang.System.getSecurityManager();
                 if (sm != null)
                     sm.checkPermission(new global::java.lang.RuntimePermission("createClassLoader"));
-                Type type = Type.GetType(name);
+                Type type = null;
+                foreach (Assembly asm in AppDomain.CurrentDomain.GetAssemblies())
```
wasabii, line 51: This seems very much wrong. It is avoiding the entire class load hierarchy, and scanning all AppDomain assemblies loaded for every class lookup? What if the hierarchy isn't configured to do so?

Also, on Core, this only uses the default assembly load context, and not any nested ones.

wasabii:
Okay. I left a few comments, while looking through it. But my overall impression I think is that it's sort of missing the point. You can't just randomly scan assemblies. There's a hierarchy to follow. Not all calls to Class.forName should find all types.

src/IKVM.Runtime/TypeWrapper.cs
DjArt: How it should be organized? There is area that can contains only unique type names?

That code was written for comparing already loaded TypeWrapper and pending to load. Were encountered some problems with loading library classes from another library.

wasabii: No. There is no such thing as unique type names. Type names can be duplicated.

DjArt:
Perhaps, first I will describe what exactly happened, perhaps there are other ways to fix it.

wasabii:
Going to close this out for inactivity.

PR #257, opened 2023-01-05
Update OpenJDK to 8u51-b34

wasabii:
Begin work on a small update of OpenJDK to see if we can pull it off a bit at a time.


wasabii:
New additions in this update: Elliptic curve classes. Implemented through .NET's EC* set of classes. Massive differences in Framework versions here. And lots of bugs.

NET461 only has partial support. It's missing the ECCurve, ECPoint, etc classes. Implemented custom ECCurve, ECPoint structs to shield the API. There's no way to export or load keys in a standard format on NET461: instead, we need to use CngKey directly, which exports BCRYPT_ECCKEY_BLOB structures. These contain magic numbers, key sizes, and uncompressed D X and Y values. So we build these structures on the way in and the way out.

NET47 is a little bit better. However both it and NETCOREAPP3.1 have a bug where you need to provide a non-empty value for X and Y when importing keys. This hack works on NET5+, though it's no longer required.

wasabii:
Because of the massive different in NET461 and NET47, we introduce NET47 builds throughout. New drops.

wasabii:
And, removing the EC stuff for now. Turns out the derive key material in .NET seems not to be exactly compatible with whatever algorithm that Java was using, and I'm not sure yet how to fix it. Removing net47 drops no longer required.

The EC classes are still present, just disabled. Needs to be addressed at a later time.

PR #278, opened 2023-01-28
Workflow update

bogatykh:
Do we need 18.04 Ubuntu to build OpenJDK?

### DIFF
--- a/.github/workflows/IKVM.yml
+++ b/.github/workflows/IKVM.yml
@@ -28,7 +28,7 @@ jobs:
         key: openjdk-build-linux-x86_64-normal-server-release--${{ runner.os }}--${{ hashFiles('openjdk/**', '!openjdk/build') }}-8
     - name: Check OpenJDK Build Stamp
       id: openjdk-build-stamp
-      uses: andstor/file-existence-action@v1
+      uses: andstor/file-existence-action@v2
       with:
         files: openjdk/build/linux-x86_64-normal-server-release/stamp
     - name: Fetch OpenJDK 7
@@ -94,7 +94,7 @@ jobs:
         key: jtreg-build--${{ runner.os }}--${{ hashFiles('jtreg/**', 'openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image', '!jtreg/build') }}-6
     - name: Check OpenJDK Test Harness Build Stamp
       id: jtreg-build-stamp
-      uses: andstor/file-existence-action@v1
+      uses: andstor/file-existence-action@v2
       with:
         files: jtreg/build/stamp
     - name: Build OpenJDK Test Harness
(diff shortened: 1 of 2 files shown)


wasabii:
We did when I originally put this together. The Lenny distro uses syscalls not available in later kernel versions, if I remember.

wasabii:
You'll have to make sure the cache is expired, or it won't even try building, though. The IKVM.yml line for the name of the NuGet cache has a number at the end that you can incremement.

bogatykh:
> You'll have to make sure the cache is expired, or it won't even try building, though. The IKVM.yml line for the name of the cache has a number at the end that you can incremement.

OK I will check it out. Anyway we will have to figure out what to do with it due to depreciation 
At least we need to update github actions to latest versions, this looks safe to me.

wasabii:
Yeah the versions are fine.

Gotta get the JDK upgraded. But it's such a slow process. The latest jdk8 builds fine.

bogatykh:
@wasabii looks like intermittent failure of unit test. Can we rerun workflow?

wasabii:
@bogatykh You'll have to make that another commit to expire the OpenJDK build cache. Else we don't know whether it actually builds.

I have it cached because it takes a significant amount of time, and it's not like we ever change it.

bogatykh:
@wasabii you may delete cached entry from web with write permissions (I don't have them).
Take a look: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#deleting-cache-entries

I can make commit later to expire cache if above instructions won't work for you.

wasabii:
You'll need to do it in your branch to generate a new cache entry for your branch (encapsulating your changes).

bogatykh:
> You'll need to do it in your branch to generate a new cache entry for your branch (encapsulating your changes).

Never done that.
Can you please verify my last commit?

wasabii:
All you got to do is increment the integer at the end.

.github/workflows/IKVM.yml
```diff
@@ -15,20 +15,20 @@ on:
...
     - name: Checkout Source
-      uses: actions/checkout@v1
+      uses: actions/checkout@v3
       with:
         submodules: recursive
     - name: Cache OpenJDK Build
       uses: actions/cache@v3
       with:
         path: openjdk/build/linux-x86_64-normal-server-release
-        key: openjdk-build-linux-x86_64-normal-server-release--${{ runner.os }}--${{ hashFiles('openjdk/**', '!openjdk/build') }}-8
```
wasabii, line 28: All you need to do is put -9 on the end of it, not change the prefix.

wasabii:
All that said, the latest Ubuntu does appear to work. Good news.

.github/workflows/IKVM.yml
bogatykh: Done. Thank you!

PR #283, opened 2023-02-06
Class writing move to ByteCode

bogatykh:


bogatykh:
@wasabii this is work in progress and is not complete. Creating PR to inform you about ongoing work.

wasabii:
Wow dude. It'll take me a bit to go over this one. But awesome. I didn't expect anybody else to go for this.

bogatykh:
@wasabii can you please suggest me Writer class design? I was using your Reader classes design as a reference and can't come up with a good design. How would you design, for example FieldWriter for FieldRecord?

wasabii:
One of the things I was going to do was introduce a Builder pattern over the records.

wasabii:
I hadn't even decided yet whether to name them Writers to be inline with Readers, or Builders, because lots of other projects use that terminology. But I had thought that modeling it sort of like System.Metadata.Reflection would be beneficial.

wasabii:
https://learn.microsoft.com/en-us/dotnet/api/system.reflection.metadata.ecma335.metadatabuilder?view=net-5.0

Obviously different concepts.

But they create a MetadataBuilder. Then invoke Add* methods on it. And metadata provides a few basic classes for data allocation, which we would use for Constant pool allocation. Like GetOrAddString(), etc.

wasabii:
At the end of the process, our ClassBuilder would have a Build() method which returned a ClassRecord. Which was immutable. And could then be written to a ClassFormatWriter.

wasabii:
@bogatykh also to make you aware, one of the users of this will be #284

bogatykh:
> One of the things I was going to do was introduce a Builder pattern over the records.

That's a good idea. I have implemented some builder class. Could you please comment if we go this way or like System.Reflection Ecma builder?

wasabii:
Obsolute. Complete rewrite of IKVM.ByteCode, with encoders and decoders. Is now used by IKVM.

PR #290, opened 2023-02-09
Upgrade OpenJDK to jdk8u60-b32

wasabii:
Contains an upgrade to JDK8u60-b32. Along with this, many other mostly forked systems have been rebased to incorporate the latest JDK code changes.

* NIO rewrite. Complete rewrite of channels, partial rewrite of files. The new design operates completely differently. No longer is it a forked copy of the Windows NIO channels, containing a "IOCP" translation layer. Instead, the DotNetChannel* classes are hand built, backed by C# code in externs. The basic abstraction is around Task based async, as that's the current direction of .NET. The internal PendingFuture class is populated from Task completion. Then various .NET APIs that are invoked are either directly invoked as their Async based component, or the TPM component is wrapped in Task.Factory.FromAsync on Framework, but the async version is used on Core.
* The ChannelGroup implementation wraps calls to the various underlying APIs using a SynchronizationContext. This way, on the .NET side, we appear to be using basic async/await, but continuations are scheduled on the ChannelGroup, as intended by the Java developer.
* Where available, Span and Memory are used. These are pretty inconsistent. Some Core methods support them. Some don't. Some Framework methods support them. Some don't. In those cases where we can, they're used. For both ByteBuffers and DirectBuffers. There are cases where we have to copy a DirectBuffer into a .NET array to invoke a .NET API that lacks Span/Memory. And there are cases where we can simply map an array across. And there are .NET APIs that can directly accept a Memory, which we can obtain from a DirectBuffer. It's all pretty inconsistent: but we use the best version for the platform we're on.
* Property loading has been moved out of VMSystemProperties.java into JVM.Properties. In hotspot, this stuff is done in native code which is part of the JVM. This reflects that model a bit better. And it lets us discard an entire class that doesn't exist in Java. A consequence of this is the template files for JVM version information was moved to IKVM.Runtime from IKVM.Java.
* The stub generator has gained some additional options.
  + --non-public-types, --non-public-interfaces --non-public-members. This allows inclusion or exclusion of these various things in the generated .JAR files. I found a number of Java test cases that actually probed pretty deep into class files to test for things.
* New Accessor framework. The idea here is to reduce the references from IKVM.Runtime to IKVM.Java so we can eventually break the cycle between the two. This is just the beginning of this. These accessors are pretty typical dynamic IL generation of field/property/method getters/setter/invokes. It might be better to actually go towards the TypeWrapper functionality of the same. But that itself needs some pretty big cleanup.
* The VFS location was altered. Instead of now being hard coded to a specific path, such as a non-existent UNC machine, and instead of being unified, different components are mounted in different places. The assembly folder is mounted in ikvm.home\assembly, which the cacerts file is directly mounted into the ikvm.home\lib\security\cacerts file. Basing the path on ikvm.home guarantees it's a valid path on all systems (after all ikvm.home has to be). And makes it look like the file is really there.
* Many OpenJDK test cases enabled that were previously failing. Many still left.
* ZipFile implementation rebased. Had an IKVM specific modification for dos times. JDK8u60 has this built in now.
* Finalizer bug fixed. The JLS says a finalizer should not run unless the java.lang.Object ctor has completed. .NET will always run the finalizer. We add a private field __init to j.l.O to track completion of this ctor, and the .NET finalizer skips execution of not set.
*


GeorgeS2019:
Current IKVM 8.4.5 is jdk8u45-b37 :-)

AliveDevil:
Fixed LambdaStackTrace (test id 8025636): https://github.com/iterate-ch/ikvm-revived/commit/8df7b16836b99270432c0a1302a97e77e9052ce0

PR #340, opened 2023-05-29
ikvmc.csproj: missing reference to runtime

Kojoley:
Without this VS does not put dlls near the exe and run command fails.

### DIFF
--- a/src/ikvmc/ikvmc.csproj
+++ b/src/ikvmc/ikvmc.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <Import Project="$(MSBuildThisFileDirectory)..\..\IKVM.deps.props" />
-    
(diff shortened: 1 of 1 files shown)


wasabii:
Those files are not supposed to be distributed alongside the tools. The tools don't consume the code within them.

Instead, the tools are passed options pointing to the runtime libraries you want to use during an invocation of the tools using the -reference and -runtime arguments.

Kojoley:
It was the easiest and straightforward way to fix 'Run' button in VS.

wasabii:
I understand. But it also results in the inclusion of IKVM.Runtime and IKVM.Java (and a few other things, such as the native libraries) as referenced assemblies in the ikvmc executable's dependency graph, which isn't correct as ikvmc does not in fact reference those assemblies.

I run ikvmc from Visual Studio by configuring the Debug arguments to include the same contents as you would include if you were running it from outside Visual Studio: --reference, --runtime, etc..

PR #351, opened 2023-06-06
Let IKVM.Net.Sdk handle java archives as opaque ikvmc conversion target

### LINKED ISSUE #326: Include JAR files in IKVM.NET.Sdk
A conversation I had on Discord with @AliveDevil resulted in this.

IKVM.NET.Sdk should support the inclusion of .jar files along side it's existing support for .class and .java files. This will assist @AliveDevil with his project. But also, it's the right place to put it. The goal of IKVM.NET.Sdk is to be a .NET Project type, which produces exactly one output assembly (like every other .NET project type), and there's nothing wrong with it being able to source the input for that production from multiple different types of files. This can be used by users to convert existing JAR files to Assemblies for the purposes of export/redistribution. IkvmReference isn't appropriate for this, because it's geared towards consuming Java libraries for the purpose of using them in an existing project, not for exporting or publishing them.

AliveDevil:
Tested in the wild (https://github.com/iterate-ch/cyberduck/compare/master...feature/winui).

IKVM builds fine (didn't see any regression).
A first converted application (CLI, as UI is somewhat complicated) is working without issue.

ProjectReference, Reference and type discovery works fine (no duplicate classes are generated).

Resolves #326

### DIFF
--- a/src/IKVM.NET.Sdk/targets/IKVM.Java.Core.targets
+++ b/src/IKVM.NET.Sdk/targets/IKVM.Java.Core.targets
@@ -113,10 +113,6 @@ Files = files.ToArray();
         </ClasspathToClasspathFiles>
     </Target>
 
-    <ItemGroup>
-        <Compile Include="$(MSBuildThisFileDirectory)_.java" />
-    </ItemGroup>
-
     <!-- Abstract target to conduct actual Java compilation. Overridden either by Command or Task implementation. -->
     <Target Name="_CompileJava" DependsOnTargets="ExportReferences;_ResolveClasspathFiles" Inputs="@(Compile);@(ClasspathFiles)" Outputs="$(_CompileJavaStampFile)">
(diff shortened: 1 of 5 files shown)


src/IKVM.NET.Sdk/targets/IKVM.Java.Core.targets
```diff
@@ -132,9 +128,9 @@ Files = files.ToArray();
         </CompileJavaDependsOn>
     </PropertyGroup>
 
-    <Target Name="CompileJava" DependsOnTargets="$(CompileJavaDependsOn)">
+    <Target Name="CompileJava" Inputs="@(Compile)" Outputs="@(Convert)" DependsOnTargets="$(CompileJavaDependsOn)">
```
wasabii, line 131: This Outputs= line breaks projects. Just noticed. Fixing in develop.

Issue #364, opened 2023-06-22
NativeUnpack Not Implemented


TheLastRar:
IKVM version : IKVM-8.5.2-image-netcoreapp3.1-win7-x64

With vanilla Minecraft mostly working, I wanted to see if modded Minecraft would work, however, forge 10.13.4 uses Pack200.
IKVM currently does not support Pack200.

I don't know if this is something that would be best included as a native library, or ported to C#

Take the following example code below, or the code listed here https://docs.oracle.com/javase/8/docs/api/java/util/jar/Pack200.html

```java
import java.io.File;
import java.io.FileOutputStream;
import java.io.IOException;
import java.util.jar.JarFile;
import java.util.jar.JarOutputStream;
import java.util.jar.Pack200;
import java.util.jar.Pack200.Packer;
import java.util.jar.Pack200.Unpacker;

public class Pack200Test {
	public static void main(String[] args) {
		// Create the Packer object
		Packer packer = Pack200.newPacker();
		
		try {
			JarFile jarFile = new JarFile("test.jar");
			FileOutputStream fos = new FileOutputStream("test.pack");
			// Call the packer
			packer.pack(jarFile, fos);
			jarFile.close();
			fos.close();

			File f = new File("test.pack");
			FileOutputStream fostream = new FileOutputStream("testExtracted.jar");
			JarOutputStream jostream = new JarOutputStream(fostream);
			Unpacker unpacker = Pack200.newUnpacker();
			// Call the unpacker
			unpacker.unpack(f, jostream);
			// Must explicitly close the output.
			jostream.close();
		} catch (IOException ioe) {
			ioe.printStackTrace();
		}
	}
}
```
You will get the following exception
```
Exception in thread "Thread-0" cli.System.NotImplementedException: The method or operation is not implemented.
        at IKVM.Java.Externs.com.sun.java.util.jar.pack.NativeUnpack.setOption(NativeUnpack.cs:60)
        at com.sun.java.util.jar.pack.NativeUnpack.setOption(Native Method)
        at com.sun.java.util.jar.pack.NativeUnpack.copyInOption(NativeUnpack.java:174)
        at com.sun.java.util.jar.pack.NativeUnpack.run(NativeUnpack.java:192)
        at com.sun.java.util.jar.pack.NativeUnpack.run(NativeUnpack.java:247)
        at com.sun.java.util.jar.pack.UnpackerImpl.unpack(UnpackerImpl.java:138)
[...]

Issue #368, opened 2023-06-25
libverify implementation


wasabii:
Initial implementation of libverify.

Issue #367, opened 2023-06-25
libjava implementation


wasabii:
Initial implementation of libjava.

Issue #366, opened 2023-06-25
libjvm implementation


wasabii:
Initial implementation of libjvm.

PR #370, opened 2023-06-25
Pack200 native implemention

wasabii:
In an effort to get Pack200 working for #364, this PR implements P2P references for ClangProjects, implements libverify #368, libjvm #366, libjava #367 and libunpack.


wasabii:
libjvm is basically hotspot. It will largely consist of hand-written functions that are not part of OpenJDK, but instead short trampolines that simply call into .NET code, and manipulate IKVM structures. This should be good in the short term, as long as libjvm is loaded up at IKVM startup.

However, long term, libjvm should more properly implement the actual libjvm API: the ability to link against it to actually host a full VM. This would potentially allow Java applications which make use of libjvm.so, etc, directly, to instead link against the IKVM VM, spawn .NET, etc.

wasabii:
fixes #364 #366 #367 #368

PR #388, opened 2023-07-17
UnixExecutableMemory -> Posix, free with munmap

wasabii:
Should have been freed with munmap. Class name was odd.


wasabii:
Shoudl have targeted main.

PR #402, opened 2023-07-29
OS X ARM64

wasabii:
Based on net6 branch. Added OSX-ARM64 output.

Deprecated NET Core 3.1 tools and images.

Tools now using NET 6 to build NET Core 3.1 libraries.


paulirwin:
Thank you for this! I am very much looking forward to this PR and #401. I attempted to do these myself and realized I was way over my head and would need to spend a lot of time learning the architecture. It will be much appreciated once this is complete!

paulirwin:
@wasabii Just wanted to say I am following this PR and I see the hard work you're doing, and that it's noticed and appreciated 😄

wasabii:
> @wasabii Just wanted to say I am following this PR and I see the hard work you're doing, and that it's noticed and appreciated 😄

Thanks.

Wish I had a Mac to run this on. The best I can do is push to CI and wait each time.

wasabii:
Well, if there was some way you could load up the branch in VS, or whatever it's called on a Mac, and run the unit tests, and then report back all the broken ones, that'd be great.

That's all I'm doing right now: disabling broken tests to get a base line, from which I can then start fixing things.

wasabii:
Also I have no idea how we're going to handle code signing.

paulirwin:
I am making progress getting it to build, but it doesn't like the JDK 17 javac I have installed. Having trouble currently getting a JDK 8 build installed via Homebrew for use under Rosetta. I'll keep digging and see if I can get one working.

wasabii:
I do not know exactly what luck you will have building the whole thing on OS X ARM64. The osxarm64 branch still has Core 3.1 building enabled, which probably isn't going to work.

paulirwin:
Gotcha, hm. I am not sure if I will be able to easily run the unit tests then locally if I can't build locally. I guess theoretically it's possible to build the binaries on one machine and then copy them over and see if Rider accepts that as a built artifact... I'll have to try that later.

Given that .NET Core 3.1 is no longer supported, is it a possibility to just drop that target altogether? Although it would still encounter a problem building for .NET FX of course. It would be great if there is a way to set a flag or use a separate msbuildproj or something that excludes .NET FX (or older Core 3.1) from the build on this platform. But I have zero idea how complicated that would be (I'm assuming the answer is "very").

wasabii:
The plan is to drop Core 3.1, yes. But that's on another branch, which will go in after OSX ARM64. So, that's unfortunate for you.

It's possible I'll have a osx-arm64 build by the end of day today. Looks like I've got almost all of the test cases excluded.

And I just pushed support for doing ad-hoc codesigning of the executables and shared libraries. Yuck.

So... soon. Maybe.

wasabii:
@paulirwin 

Develop branch has OSX-ARM64 merged in. And has built and generated a package in the GitHub private repository.

Would you be interested in testing?

Most of the unit tests ran on CI/CD, on OSX-X64. And I got the signing in place. And I signed up for a quick Scaleway M1 instance for 30 minutes to run the OpenJDK test suite. It mostly all ran on ARM64.

So, yeah, I've never directly tried all of this on OSX-ARM64. So it might not work. But it also might work.

Give it a go!

wasabii:
@BiologyTools You too.

wasabii:
Eh. Already found a bug. Bad path in the MSBuild files for the SDK tools. So, IkvmReference will fail.

Fixed. New thing pushed. Another 8 hours until build done. Heh.

In the meantime you can probably plop this into your project file at the bottom to fix it:

```
    <ItemGroup>
        <IkvmCompilerToolPath Include="$(NuGetPackageRoot)\ikvm.msbuild.tools.runtime.osx-arm64\8.7.0-dev.268\ikvmc\net472\osx-arm64\" TargetFramework="net472" RuntimeIdentifier="osx-arm64" />
        <IkvmCompilerToolPath Include="$(NuGetPackageRoot)\ikvm.msbuild.tools.runtime.osx-arm64\8.7.0-dev.268\ikvmc\net6.0\osx-arm64\" TargetFramework="net6.0" RuntimeIdentifier="osx-arm64" />
        <IkvmExporterToolPath Include="$(NuGetPackageRoot)\ikvm.msbuild.tools.runtime.osx-arm64\8.7.0-dev.268\ikvmstub\net472\osx-arm64\" TargetFramework="net472" RuntimeIdentifier="osx-arm64" />
        <IkvmExporterToolPath Include="$(NuGetPackageRoot)\ikvm.msbuild.tools.runtime.osx-arm64\8.7.0-dev.268\ikvmstub\net6.0\osx-arm64\" TargetFramework="net6.0" RuntimeIdentifier="osx-arm64" />
    </ItemGroup>
    ```

wasabii:
That last build thing is fixed in .269

paulirwin:
Will do, thanks! What is the best way to try this out? Download the nuget artifact from the build (I'll probably wait until .269 is complete) and install locally into my project?

BiologyTools:
> @BiologyTools You too.

Unfortunately I don't have an arm64 Mac to test things on. I only have a x64 mac. But thanks for the mention and the great work you are doing. I bet many biologists will download the bioformats library in NET6 once IKVM 8.7.0 is released. I could try and ask for help with testing on the bioformats image.sc forum.

wasabii:
@paulirwin  Add the Github private repo as a nuget source.

paulirwin:
Ah gotcha, my mind was thinking "Git repo" when you said that and I was assuming you meant the build artifacts. Will do!

paulirwin:
I'm not sure what the problem is but I can't seem to install from the GitHub NuGet feed. I created a new PAT with the `read:packages` permission, and added what I think is the right source url (https://nuget.pkg.github.com/ikvmnet/index.json) because it shows the 8.7.0 prerelease package successfully in Rider, but when I try to install/restore from either Rider or the `dotnet` CLI I get a 401 error, saying my token isn't valid.

Is that because I'm not a contributor to the repo or something? Let me know if you have any guidance; this is the first time I've tried this with a public GitHub project that I'm not a member of. Thanks!

paulirwin:
My apologies for the delay on this, but I saw that the prerelease packages were published, and I was able to update my JavaToCSharp project to use these in a branch, and it works great on macOS Arm64. I can even now run the Avalonia GUI directly without having to cross-compile from an x64 machine. I'm extremely grateful for this effort! Branch: https://github.com/paulirwin/JavaToCSharp/tree/feature/ikvm-8.7

Issue #432, opened 2023-10-20
Windows ARM64 Support (Runtime/SDK) (Framework)


wasabii:
Duplicated from #197, same restrictions apply. Framework support missing due to our lack of PE generation.

PR #469, opened 2024-01-15
Use System.Reflection.Metadata to write assemblies in IKVM.Reflection. Enable win-arm64 Framework tools and images.

wasabii:
Changes the IKVM.Reflection code to write PE files using System.Reflection.Metadata. This lets us capitalize on that code instead of maintaining our own. It also enables features that our PE generation did not have: Windows ARM64 Framework executables.

Thus the ARM64 builds are also enabled.

We still do signature writing, and IL generation, using our custom code. We also still READ assemblies using the original code. These two can be done next.


wasabii:
fixes #432

PR #494, opened 2024-03-11
Add IgnoresAccessChecksToAttribute into generated assemblies

wasabii:
… their references. The idea here is to solve the package-private issue. Since Java allows package-private access between JARs, and IKVM generally emits a different assembly per-JAR, we need a way to allow these across. One options was to change pacakge-private to public. This solution however might work.


wasabii:
This can't work. Violates the JVM spec. Need to think more.

PR #501, opened 2024-03-27
Platform Specific Base Libraries

wasabii:
Implementing support for platform specific IKVM.Java. The three different versions will be distributed and each loaded based on the runtime OS or target runtime.

Added a ModuleInitializer annotation similar to Roslyn's. Not explicitely used here, but useful nonetheless.


wasabii:
Closing this one in favor of the more complete #521

PR #562, opened 2024-07-23
Add static compiler-flag

AliveDevil:


wasabii:
My only hesitation at this point is around the name of the MSBuild property: CompileStatically doesn't strike me as correct.

AliveDevil:
How about `DisableDynamicTypeResolution`?

wasabii:
Is just 'Static' an option?

AliveDevil:
Could be, but just "Static" for me isn't a descriptive property.
Sure, it is used in IKVM.NET.Sdk, but that is also used when importing `IKVM`, and what does "Static" mean in a csproj then?
So, if you want to have a 1:1 mapping of MSBuild property to ikvmc arg, prefixing it with Ikvm or Ikvmc would be ideal - signalling that this is exclusively used for IKVM.

Suggesting `Static`, `IkvmStatic`, `IkvmcStatic`.

wasabii:
What do you mean importing IKVM? You mean IKVM reference? That doesn't use global properties.

AliveDevil:
Ah, didn't know that. Thought that the IkvmReference would also use the IKVM.NET.Sdk-stuff for invoking ikvmc.

wasabii:
The IkvmCompiler and IkvmExporter task are shared, but that's about it.

I like the more verbose name better (DisableDynamicTypeResolution). I tried to think it through. Problem is, -static itself is sort of a hack right now on ikvmc.... and includes more than just dynamic type lookup. But also a codegen option to disable refemit in various circumstances. It's not exactly strictly defined as to what 'static' means here. And I can imagine us adding more options here too: NAOT scenarios might want to limit certain things, but allow other things.

I don't like it as an option at all on ikvmc. But, it's there. And it's going to have to stay.

I figure if "Static" doesn't overlap with any javac options (and it doesn't). And we can add a similar "Static" metadata value to IkvmReference.... that's probably fine. For now.

AliveDevil:
Ack.
Expose coarse Static property to enable current ikvmc StaticCompiler, and gradually remove Static-behavior with finer grained properties in the future.

PR #567, opened 2024-07-25
Handle MSBuild properties NoWarn, TreatWarningsAsErrors, WarningsAsErrors

AliveDevil:


wasabii:
So I remember the issue with why I didn't complete this. We also run javac. It also has warnings, warn as error, etc. need that to work also.

AliveDevil:
Think I mixed some things up in the commit graph here. Need to clean that up on Monday.

AliveDevil:
> why I didn't complete this. We also run javac.

I suggest to make the javac-part for error reporting a second PR - as the ikvmc build warnings could benefit consuming projects now.

wasabii:
Yeah but the options will overlap. That's what I mean

wasabii:
Okay. Here's what I think.

Predominately people are going to be using IKVM.NET.Sdk to build Java code. Not just to assembly class files or JARs like you're mostly doing. It's a Java-language project type primarily. So, they'd expect all the top level properties to be written from that perspective. Hence, WarningsAsErrors, etc, predominatley, should be thought of from that perspective: people are mostly going to be putting Java warning codes into them. To silence Java-language warnings (or enable them).

But, javac doesn't have error codes like C# does. Instead, it has -Xlint:+foo, where you can enable or disable a linter. Which is basically the same thing. Except it has no indepdent way to treat it's output as an error. It's always a warning, and then globally you can set -Werror to turn all warnings into errors. So it doesn't quite match up with the MSBuild properties that make sense from a C# perspective or a IKVM perspective.

And an additional thing to throw into the mix: we are not the only project-type that acts as a composite of other tools, each with their own warning/error thing. NuGet pack, for instance, can generate warnings. But then there's also Razor. And a bunch of other stuff. Each of which supports some sort of messing around with this same property.

I built a sample project and messed with NoWarn, WarningsAsErrors, etc. What I found was that whatever you stick into it gets passed into all these tools. Same convention between all the MS tools. The tools all seem to operate the same way, also: if they are passed a warning value (1234), and they do not support it, they silently ignore it. They also all support optional prefixes: CS0001, NU0001, etc. So the user can set one of these options, the MSBuild script can pass it to everything, but only the tool it applies to will respect it, and nothing else will break.

So, as long as IKVMC behaves the same way.

Looks like it doesn't.

When processing these options, as far as I can tell, it only supports a LongText version and an int version. Where the LongText version is just the enum name, for instance.

EmittedIllegalAccessError = 112,

So, -nowarn:112 works. -nowarn:EmittedIllegalAccessError works. But -nowarn:IKVMC0112 does not seem to work. It does strip leading zeros, but not the leading IKVMC prefix.

It also prints out IKVMC0112 in a few places. So, it prints out a format it doesn't accept? Seems so.

So, the user setting `<WarningsAsErrors>1000</WarningsAsErrors>` would be what a user would do to disable a IKVMC warning.

Problem is that overlaps with NU1000: https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1000

So, at least if I'm correct here, the user has no way to treat the IKVM warning as an error, without accidently treating the NuGet warnings as an error. And no way to really ignore one or the other. I didn't test this btw, just reading the code.

I think we need to have IKVMC process the IKVMC prefix on the numbers to solve this. That way we can behave like most .NET tools, and just pass all warning/error IDs to everything and let the tools sort it out.

As to the javac options: I'm not sure yet. Part of me wants to think through inverting the Xlint options so they BEHAVE like NoWarn, WarningsAsErrors, etc. After all, we don't actually run javac. We run the JavaCompiler task, which gives us a bit more flexibility, I think. We might be able to actually change how Xlint works to make it more .NETish.

I think we can ignore that for now though, as long as we stick to the rule that we behave the same as other .NET tools: we accept any ID, we check ALPHAprefix if it exists, and ignore if it's an invalid number. So I think we might need to fix that one issue in ikvmc.

AliveDevil:
I see. Going to amend this with parsing `IKVMC`-prefixed errors. So `<NoWarn>1000;IKVMC1000;WhateverEnumNameIsUsedInIKVM</NoWarn>` would be considered identical.
And checking whether there are any places that don't adhere to the message handlers.

AliveDevil:
Waiting for forked dev-build to verify function.

AliveDevil:
NoWarn worked … but WarningsAsErrors didn't.

src/IKVM.Tools.Importer/StaticCompiler.cs
```diff
@@ -240,250 +240,7 @@ internal void IssueMessage(Message msgId, params string[] values)
 
         internal void IssueMessage(CompilerOptions options, Message msgId, params string[] values)
         {
-            if (errorCount != 0 && msgId < Message.StartErrors && !options.warnaserror)
```
AliveDevil, line 243: This was a literal copy of IkvmImporterInternal, forwarding to there.

src/IKVM.Tools.Runner/Importer/IkvmImporterLauncher.cs
```diff
@@ -299,8 +299,9 @@ public async Task<int> ExecuteAsync(IkvmImporterOptions options, CancellationTok
                 await LogEvent(IkvmToolDiagnosticEventLevel.Debug, "Executing {0} {1}", cli.TargetFilePath, cli.Arguments);
 
                 // send output to MSBuild
-                cli = cli.WithStandardErrorPipe(PipeTarget.ToDelegate(i => LogEvent(IkvmToolDiagnosticEventLevel.Error, i)));
-                cli = cli.WithStandardOutputPipe(PipeTarget.ToDelegate(i => LogEvent(IkvmToolDiagnosticEventLevel.Debug, i)));
+                var logUnknown = PipeTarget.ToDelegate(i => LogEvent(IkvmToolDiagnosticEventLevel.Unknown, i));
```
AliveDevil, line 302: Handle stdout & stderr using Unknown log level (i.e. let MSBuild figure it out)

src/IKVM.Tools.Importer/IkvmImporterInternal.cs
```diff
@@ -1526,6 +1510,35 @@ DiagnosticLevel.Warning when options.errorWarnings.ContainsKey(key) || options.e
...
+                    parse = w.Substring(prefixStart, contextIndex - prefixStart);
+                }
+                else
+                {
+                    parse = w.Substring(prefixStart);
+                }
+
+                if (!int.TryParse(parse, out var intResult))
+                {
+                    // NamedResults aren't supported
```
AliveDevil, line 1535: This would've required Reflection, or the .tt-file needs to emit a huge switch for parsing.

wasabii: Yeah, I removed worrying about the name versions, because none of the other tools support it that I could find (csc, etc).

wasabii: SuppressWarnings will probably go away eventually too, but I wasn't exatly sure under what circumstances it was used. It's used, I think, to prevent duplicate warnings/errors from appearing. But it's such a round-about way to go about it.

PR #576, opened 2024-08-15
`Assembly.Location` might return an empty string.

Fabian-Schmidt:
`Assembly.Location` might return an empty string.
`Path.GetDirectoryName` on Full Framework will then fail with `The path is not of a legal form.`.

### DIFF
--- a/src/IKVM.Runtime/LibIkvm.cs
+++ b/src/IKVM.Runtime/LibIkvm.cs
@@ -218,7 +218,7 @@ namespace IKVM.Runtime
                 return h1;
 
             // start looking at assembly path, or path given by environmental variable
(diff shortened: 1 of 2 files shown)


Fabian-Schmidt:
Hi, I signed the CLA, but made first commit on my new Dev VM without correct setup git config and therefore wrong Name & EMail is associated with the commit.
To fix it I would have to recreate the commits and recreate the PR. Do you want me todo this or is it acceptable in this state?

AliveDevil:
You could also force push the changed commits.

wasabii:
Hmm. I think I would like you to fix this. Just amend the history and force push. I think the risk is low, but it is tied to a legal thing and so might be important (to some person in the future).

Issue #580, opened 2024-08-22
Adding User setting before initialization


except11:
Hi, I was trying to use User settings to override ikvm.home.root since I saw in code IKVM checking it first

I am doing it like this
IKVM.Runtime.JVM.Properties.User.Add("ikvm.home.root", "ikvm-root");

But to do it system tries to construct static objects and tries to get home path before I can set it

System.TypeInitializationException: The type initializer for 'Internal' threw an exception.
 ---> IKVM.Runtime.InternalException: Could not locate ikvm home path.
   at IKVM.Runtime.JVM.Properties.GetHomePath()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at IKVM.Runtime.JVM.Internal..cctor()
   --- End of inner exception stack trace ---
   at IKVM.Runtime.JVM.Init()
   at IKVM.Runtime.RuntimeInit.Init()
   at .cctor()

I managed to solve my problem by adding PrivateAssets="buildTransitive" to reference and copying bunch of assets myself, since I can't just replace ikvm.properties - its added to build forcefully, and I want my own version.

Maybe add ikvm.user.properties for this purpose or something? Currently it looks you can customize the path in different ways but none of them actually work as is

wasabii:
You can just use the build script to lay down your own ikvm.properties. But yeah, we can't do any code-based dynamic lookup of this because module inits happen before everything.  On Framework, app.config works since it's loaded by the framework. We could find a similar thing for Core.

wasabii:
Replacing ikvm.properties shouldn't be that hard though. You should probably just be able to override the target that lays it down.

except11:
Yes, I did it like this
  <Target Name="ReplaceIkvmProperties" AfterTargets="GetDefaultIkvmPropertiesFile">
    <ItemGroup>
      <None Remove="@(None)" Condition="'%(Extension)'=='.properties'" />
      <Content Remove="@(Content)" Condition="'%(Extension)'=='.properties'" />
      <None Include="ikvm.properties">
        <CopyToOutputDirectory>Always</CopyToOutputDirectory>
      </None>
    </ItemGroup>
    <Message Text="Replaced ikvm.properties @(None)" Importance="high" />
  </Target>
  
[...]

PR #590, opened 2024-09-08
Search for IKVM.Home and IKVM.Home.Root in AppContext

wasabii:
Should allow configuration of these variables through RuntimeHostConfigurationOption.


wasabii:
Should provide an option for #580

PR #592, opened 2024-09-13
New CoreLib Symbols support

wasabii:
Provides IKVM.CoreLib.Symbols, which contains facades for all of the reflection-like entities used: ITypeSymbol, IAssemblySymbol, etc. These interfaces are then implemented for System.Reflection and IKVM.Reflection. This allows us to start untying the Runtime from the IFDEFs that change which reflection interface is used.

There some overhead here since we're wrapping the entire APIs with interfaces. Virtual dispatch. But, we might be able to actually eak out some performance improvements with caching.

Additionally, now we have an API surface we can start evolving differently than System.Reflection. Once we start narrowing the interface down, removing the methods we do not use, etc, we might be able to add higher level more focused methods that we could then go directly to System.Reflection.Metadata for, ditching IKVM.Reflection completely.


wasabii:
Some obvious first things might be to eliminate the array copies that come from GetMethods, GetTypes, etc. System.Reflection mandates that these calls return a new array each time. And the caller is allowed to modify it. And IKVM does this in a few places. But, we could convert these to ImmutableArrays, and cache the results, reducing the number of array copies made, and let IKVM code make copies as it requires.

PR #597, opened 2024-09-19
Load IKVM.Properties from RuntimeHost Configuration

AliveDevil:
Allows for specifying system properties like `java.net.preferIPv4Stack` before JVM is initialized when consuming IKVM as a runtime library.

Usage:
```
<ItemGroup>
    <RuntimeHostConfigurationOption Include="IKVM.Properties" Value="java.net.preferIPv4Stack" />
    <RuntimeHostConfigurationOption Include="java.net.preferIPv4Stack" Value="False" />
</ItemGroup>
```

IKVM.Properties is semicolon separated.

### DIFF
--- a/src/IKVM.Runtime/JVM.Properties.cs
+++ b/src/IKVM.Runtime/JVM.Properties.cs
@@ -297,6 +297,19 @@ namespace IKVM.Runtime
                 }
 #endif
 
+#if NET
+				if (AppContext.GetData("IKVM.Properties") is { } ikvmPropertiesContext && !string.IsNullOrWhiteSpace(ikvmPropertiesContext))
+				{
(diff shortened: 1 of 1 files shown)


src/IKVM.Runtime/JVM.Properties.cs
```diff
@@ -297,6 +297,19 @@ static void InitSystemProperties(Dictionary<string, string> p)
                 }
 #endif
 
+#if NET
+				if (AppContext.GetData("IKVM.Properties") is { } ikvmPropertiesContext && !string.IsNullOrWhiteSpace(ikvmPropertiesContext))
```
AliveDevil, line 301: File is mixed Tabs/Spaces.
Aligned with method definition (Tabs).

wasabii:
I might change this around to just have it all inside one comma separated option.

PR #625, opened 2024-11-26
Fixes "Parameter "message" must not be null" in IKVMCompiler task

AliveDevil:


AliveDevil:
Drafted, amending with IkvmCompiler not correctly suppressing warnings.

PR #716, opened 2026-02-28
Update to jdk8u482-b08

wasabii:


wasabii:
@copilot Can you either fix or disable these broken tests? And summarize which tests.

wasabii (commented):
@copilot Bro you can't just disable Framework completely. Framework on Windows is fine.

Issue #572, opened 2024-08-02
ServiceLoader.load() not enumerating services definition


AliveDevil:
Repro: [serviceload.zip](https://github.com/user-attachments/files/16471477/serviceload.zip)

Actual behavior: UnsupportedOperationException thrown, with message "No scheme SIV_GCM".

Expected: Service Loader finds CryptorProvider service

AliveDevil:
Setting
```
[assembly: CustomAssemblyClassLoader(typeof(AppDomainAssemblyClassLoader))]
```

In the `app.Launcher` resolves the issue, which is counter-intuitive.

wasabii:
I think we resolved this by talking about ClassLoader stuff.

Issue #323, opened 2023-05-12
Java Service is not Loaded for a Exe project


realmoieen:
I've created a Library type project called **ProjectLib** which covert the fat jar that contains the **META-INF.services** and output a dll file. In this Library project I added a wrapper class to jar classes So that I can use this C# Lib project into another **EXE** type project. But when I run the **ProjectExe** an Exception occurred.

`
Exception thrown: 'java.util.NoSuchElementException' in IKVM.Java.dll
An unhandled exception of type 'java.util.NoSuchElementException' occurred in IKVM.Java.dll
Unhandled Exception: java.util.NoSuchElementException
   at java.util.ServiceLoader.LazyIterator.nextService()
   at java.util.ServiceLoader.LazyIterator.next()
   at java.util.ServiceLoader.1.next()
   at com.ascertia.test.service.HelloServiceLoader.loadMyHelloService()
   at Com.Ascertia.Test.API.API.sayHelloFromNet(String name)
   at Com.Ascertia.Test.API.Exec.CallAPIFromExe.Main(String[] args)
`
It seems that the services are not loaded when the code is called from exe file.

But When I change my **ProjectLib** type to **EXE** then the META-INF/services are found and the java code is successfully executed. But I don't want to use this approach.

My approach is 
**ProjectLib** use to convert jar to dll and contain some wrapper C# classes 
**ProjectExe** an executable program using **ProjectLib.dll** and **ConvertedJar.dll** as class path dependency

I've created a demo project to reproduce the issue you can check here [Jar-To-Dll-IKVM](https://github.com/realmoieen/Jar-To-Dll-IKVM)

wasabii:
Sounds ClassLoader related. However it's going to depend version much how that Java code is attempting to scan for the file.

By default each converted Assembly is loaded within it's own ClassLoader, of type AssemblyClassLoader, which is smart enough to trace .NET Assembly references when delegating getResources or findClass calls to the AssemblyClassLoader of those other  assemblies.

realmoieen:
I'm using Azul Zulu JDK11 to compile Jdk1.8 target compatible jar file and IKVM 8.4.5 using with net461.

If sounds class loader issue then how this jar works fine when I directly add IkvmReference in an EXE project and use java
classes. 

Does I need to explicitly set JAVA_HOME environment variable or IKVM.Java.dll handle it self.
[...]

PR #753, opened 2026-09-13
Embed an ikvm.exports resource for IkvmReference items

wasabii:
## Problem

`RuntimeAssemblyClassLoader` builds its delegation set from the `ikvm.exports` resource, falling back to the assembly reference table when there is none. `ikvmc` writes that resource and names every assembly it was handed, so a jar that declares a dependency it never calls into still delegates to it at runtime.

Nothing does the same for the project that declares the `IkvmReference`. A compiler only emits an assembly reference for an assembly whose types the compiled code actually uses, so a reference reached purely by name at runtime — the service locator pattern being the usual case — leaves no trace in the output assembly, and the class loader has nothing to follow. This is the counter-intuitive part of #572, where the fix turned out to be putting `[assembly: CustomAssemblyClassLoader(typeof(AppDomainAssemblyClassLoader))]` on the app launcher so the loader would scan the whole AppDomain instead. It is also the case described in #323: *"unless your Assembly A has a reference to Assembly B, the ClassLoader for Assembly A will never find Assembly B."*

Runtime-scope Maven artifacts are the sharpest version of it. They are marked `ReferenceOutputAssembly=false`, so they are deployed but never handed to the compiler at all — there is no reference for it to emit even in principle.

## Change

Write the same resource for those projects. `GenerateIkvmAssemblyExports` in `IKVM.IkvmReference.targets` runs before `PrepareResources`, hands `@(IkvmReferenceItem->'%(CachePath)')` to a new `IkvmWriteExportsFile` task, and embeds the result with `LogicalName="ikvm.exports"`. Because it reads `@(IkvmReferenceItem)`, `MavenReference` is covered for free through `GetMavenIkvmReferenceItemsDependsOn`.

The list is exactly the `IkvmReference` items and nothing else — not `@(IkvmFrameworkReference)`, not the reference graph RAR hands the compiler.

**Ordering.** `IkvmReferenceItemPrepare` emits in build order, dependencies ahead of the items that pulled them in. A lookup wants the opposite, so the exports are written in reverse: a class carried by two artifacts resolves to the one the project declared rather than to whichever copy sits deepest, the way the nearest declaration wins a Maven class path.

**Guards.** Skipped for `$(Language) == 'Java'` — that assembly comes from `ikvmc`, which writes the resource itself. Skipped when there is no `IkvmReferenceItem`, and skipped when the items resolve to no readable assembly, so a project that declares nothing is byte-identical to before. `IkvmGenerateAssemblyExports=false` opts out.

The file is left untouched when its contents have not changed, so an unchanged reference set does not force a recompile.

## Runtime

`ikvm.exports` used to *replace* the assembly references rather than add to them. That was harmless while only `ikvmc` wrote it — the assemblies it names are a superset of the references it emitted — but it would now drop delegation to whatever the compiler did emit. `DoInitializeExports` merges the two, exports first so the declared order still wins, deduplicated case-insensitively on full name. No behavioral change for existing `ikvmc` output.

## Notes

- Provided-scope Maven artifacts are in the list but are not deployed (`Private=false`), so each costs one `Assembly.Load` that fails and is then cached as a cleared name. Filtering the list on `%(Private)` would drop them; left in for now, since excluding them breaks the case where the assembly is supplied by some other route.
- Non-Java assemblies are deliberately not filtered out of delegation anywhere. `AssemblyLoader.DoLoad` branches on `isJavaModule[i]`, and the false branch is what makes `cli.*` types resolve through a reference.
- `IKVM.Maven.Sdk` pins `IKVM 8.15.0`, so the end-to-end assertion belongs there once that reference moves past this change.

## Testing

Unit tests for the writer cover the format, ordering, reversal, deduplication, missing and non-assembly inputs, the write-only-when-different behavior, and the empty cases.

CI is green on `894b846`. Everything covering this change passed on the first attempt: all five `Test (IKVM.MSBuild.Tasks.Tests)` variants — `net472:win-x64`, `net8.0:win-x64`, `net8.0:linux-x64`, `net10.0:win-x64`, `net10.0:linux-x64` — plus `Test (IKVM.MSBuild.Tests:net10.0:win-x64)` and `Test (IKVM.NET.Sdk.Tests:net10.0:win-x64)`, which build sample projects and so exercise `GenerateIkvmAssemblyExports` end to end.

Two unrelated jobs failed on the first attempt and passed unchanged on a re-run of the same commit: `Test (IKVM.CoreLib.Tests:net472:win-x64)`, which hung in `Setup WSL` before any test ran, and `Test (IKVM.OpenJDK.Tests?TestPartition=9:net8.0:win-x64)`. See the comments below.


wasabii:
`Test (IKVM.CoreLib.Tests:net472:win-x64)` failed on `894b846`, and it is not this PR's failure.

The job never reached the tests. It hung in step 18, **Setup WSL**, which went in progress at 21:49:45Z and was still there when the job was killed at 22:34:51Z — 45 minutes. `Get Test Name`, `Download Tests`, `Restore Tests` and `Execute Tests` all stayed pending, so no test body executed and nothing from this branch was even downloaded to the runner.

Nothing in the diff can reach it either: it touches no workflow or runner configuration, `IKVM.CoreLib.Tests` declares no `IkvmReference` so the new `GenerateIkvmAssemblyExports` target does not run for it, and the `RuntimeAssemblyClassLoader` change is in `IKVM.Runtime`, which `IKVM.CoreLib` does not reference.

It is not red on the base branch: the same job passed on `73304ec`. That run failed on two different jobs, `Test (IKVM.Reflection.Tests:net8.0:win-x64)` and `Test (IKVM.OpenJDK.Tests?TestPartition=8:net8.0:win-x64)`, neither of which has gone red here.

There is no fix to port — the setup action hung, it is not a code defect. I will re-run the job; GitHub currently refuses (`403 This workflow is already running`) because the rest of the matrix is still going, so it has to wait for the run to finish. This is the one re-run, and a second failure gets treated as real.

Meanwhile the jobs that actually cover this change, `Test (IKVM.MSBuild.Tasks.Tests:net8.0:win-x64)` and `Test (IKVM.MSBuild.Tasks.Tests:net8.0:linux-x64)`, have both passed. The net472 and net10.0 variants are still queued.

---
_Generated by [Claude Code](https://claude.ai/code)_

wasabii:
Second failure on `894b846`: `Test (IKVM.OpenJDK.Tests?TestPartition=9:net8.0:win-x64)`. Unlike the CoreLib one, I cannot rule this out as not being this PR's, so recording what I have.

**What happened.** The job ran 00:15:20Z → 01:07:22Z and died with `Execute Tests` still in progress; `Archive Test Results` and `Upload Test Results` never ran, so there is no result artifact. No `##[error]` and no failing test in the log: every result line I can retrieve reports `Passed`. The retrievable tail covers about 25 seconds of jtreg report output in `java/awt/Headless` and `java/awt/im/Headless` — the suite is verbose enough that 4000 lines is 25 seconds, so I cannot see the whole run.

It is not a job timeout. The workflow's limits are 120–240 minutes, and `TestPartition=9:net472:win-x64` passed on this same commit after 75 minutes.

**Why I am not calling it mine, and why I am not calling it not-mine.** Partition 9 passed on the base commit `73304ec`, so this is not inherited the way `TestPartition=8:net8.0:win-x64` and `IKVM.Reflection.Tests:net8.0:win-x64` are — those are red on main right now and have not gone red here. And the diff does touch `RuntimeAssemblyClassLoader`, which every jtreg test exercises, so "the diff cannot reach it" is not available to me here.

Reading the change adversarially against a hang:

- The added `Assembly.GetReferencedAssemblies()` call sits inside the same `lock (this)` in `DoInitializeExports` under which the pre-existing `else` branch already makes the identical call, so it puts no new kind of API under that lock.
- The merged list is only ever appended to, and neither `FindReferenced` nor `FindDelegateResources` recurses into a delegate's own delegates, so a longer list cannot loop or blow up.
- For `IKVM.Java` the merge never runs at all: `BootstrapClassLoader` passes fixed references, and `references ??=` short-circuits before the merge is evaluated.

That is an argument, not evidence. `java/awt/**` is also the single most excluded area in `ExcludeList.txt`, and AWT on Windows runners is a well-known source of hangs.

**What blocks fixing it.** There is no identified defect to fix and no failing test named, so there is nothing to patch; and I cannot reproduce locally, as this environment has no .NET SDK. The next step is the re-run, which is owed and unspent — `rerun_failed_jobs` covers both failed jobs in one call and still returns `403 This workflow is already running` while the matrix is in flight. If partition 9 reproduces on the re-run I will treat it as real and root-cause it rather than re-running again.

---
_Generated by [Claude Code](https://claude.ai/code)_

wasabii:
Both failures cleared on the re-run of the identical commit `894b846`. CI is green: 96 of 97 jobs passed and `Release` is the only one still running.

- `Test (IKVM.CoreLib.Tests:net472:win-x64)` — passed in 4m19s, against 48 minutes hung in `Setup WSL` on the first attempt. That settles it as the environment hiccup described in the earlier comment.
- `Test (IKVM.OpenJDK.Tests?TestPartition=9:net8.0:win-x64)` — passed in 45m39s.

Correcting myself on the second one: I said above that I could not rule it out as being this PR's. Passing on a re-run of the same commit, with no change in between, rules it out. It is flaky, consistent with the rest of the evidence — on `net8.0:win-x64` partition 9 was the only one of sixteen that failed, and the partitions trade places between runs, with partition 8 red on `main` at `73304ec` but green here.

Coverage for the change itself was green throughout, on the first attempt, with no re-runs: all five `Test (IKVM.MSBuild.Tasks.Tests)` variants (`net472:win-x64`, `net8.0:win-x64`, `net8.0:linux-x64`, `net10.0:win-x64`, `net10.0:linux-x64`), which run the writer's unit tests, plus `Test (IKVM.MSBuild.Tests:net10.0:win-x64)` and `Test (IKVM.NET.Sdk.Tests:net10.0:win-x64)`, which build sample projects and so exercise `GenerateIkvmAssemblyExports` end to end.

---
_Generated by [Claude Code](https://claude.ai/code)_

